Issue #1, opened 2018-04-05
Create report output format


SolomonShorser-OICR:
Create consistent, structured, report output format and Guanming and Fred will use it.

PR #6, opened 2018-04-09
Feature/unify existing code

SolomonShorser-OICR:
- Adds QACommon which contains Report interface.
 - Adds release-qa and release-qa-distribution projects which allow all code to be built into a single jar.


SolomonShorser-OICR:
Fixes https://github.com/reactome/release-qa/issues/1
Fixes https://github.com/reactome/release-qa/issues/2

PR #15, opened 2020-05-14
DEV-1787 -- check for new, unreviewed regulatedBy instance in RlEs

cookersjs:
Small PR that adds QA check for finding ReactionlikeEvents that have received a new 'regulatedBy' instance between releases, but that has not yet reviewed.

This is added to the SliceQACheck annotation

### DIFF
--- a/src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
+++ b/src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
@@ -91,39 +91,18 @@ public class NewRegulationChecker extends AbstractQACheck implements ChecksTwoDa
     }
 
     /**
-     * Adapted from private InstanceUtilities.compareAttributes method. Takes the two 'regulatedBy' lists from the
-     * current and previous versions of a ReactionlikeEvent, and compares their contents, returning false if they differ.
+     * Takes the two 'regulatedBy' lists from the current and previous versions of a ReactionlikeEvent, and
+     * compares their contents, returning false if they differ.
      * @param regulatedByInstancesCurrent List<GKInstance> -- Contents of currentRlE's 'regulatedBy' attribute
      * @param regulatedByInstancesPrevious List<GKInstance> -- Contents of previousRlE's 'regulatedBy' attribute
      * @return boolean -- true if lists are equal, false if not.
      */
     private boolean isSameRegulatedByValues(List<GKInstance> regulatedByInstancesCurrent, List<GKInstance> regulatedByInstancesPrevious) {
-        List<GKInstance> list1Copy = new ArrayList(regulatedByInstancesCurrent);
-        List<GKInstance> list2Copy = new ArrayList(regulatedByInstancesPrevious);
-        GKInstance instance1;
-        GKInstance instance2;
-        for (Iterator it = list1Copy.iterator(); it.hasNext();) {
-            instance1 = (GKInstance) it.next();
-            for (Iterator it1 = list2Copy.iterator(); it1.hasNext();) {
-                instance2 = (GKInstance) it1.next();
-                if (instance2.getDBID().equals(instance1.getDBID())) {
-                    it1.remove();
-                    it.remove();
-                    break;
-                }
-            }
-        }
-        return equivalentLists(list1Copy, list2Copy);
-    }
-
-    // Even if Lists are the same size, it could mask unequal contents.
-    // This checks that the Lists and their contents are in fact equal.
-    private boolean equivalentLists(List<GKInstance> list1, List<GKInstance> list2) {
-        if (list1.size() != list2.size()) {
+        if (regulatedByInstancesCurrent.size() != regulatedByInstancesPrevious.size()) {
             return false;
         }
-        Set<Long> list1DbIds = getInstanceListDBIDs(list1);
-        Set<Long> list2DbIds = getInstanceListDBIDs(list2);
+        Set<Long> list1DbIds = getInstanceListDBIDs(regulatedByInstancesCurrent);
+        Set<Long> list2DbIds = getInstanceListDBIDs(regulatedByInstancesPrevious);
         return list1DbIds.equals(list2DbIds);
     }
 
--- a/src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
+++ b/src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
@@ -75,19 +75,18 @@ public class NewRegulationChecker extends AbstractQACheck implements ChecksTwoDa
         // If previousRlE does not exist, we ignore it since that does not pertain to this particular QA check.
         if (previousRlE == null) {
             return false;
-        } else {
-            // Compare contents of the 'regulatedBy' attribute of the current and previous versions of the RlE.
-            boolean sameRegulatedBy = isSameRegulatedByValues(
-                    currentRlE.getAttributeValuesList(ReactomeJavaConstants.regulatedBy),
-                    previousRlE.getAttributeValuesList(ReactomeJavaConstants.regulatedBy)
-            );
-
-            int currentRlEReviewed = currentRlE.getAttributeValuesList(ReactomeJavaConstants.reviewed).size();
-            int previousRlEReviewed = previousRlE.getAttributeValuesList(ReactomeJavaConstants.reviewed).size();
-
-            // The actual QA check
-            return !sameRegulatedBy && currentRlEReviewed == previousRlEReviewed;
         }
+        // Compare contents of the 'regulatedBy' attribute of the current and previous versions of the RlE.
+        boolean sameRegulatedBy = isSameRegulatedByValues(
+                currentRlE.getAttributeValuesList(ReactomeJavaConstants.regulatedBy),
+                previousRlE.getAttributeValuesList(ReactomeJavaConstants.regulatedBy)
+        );
+
+        int currentRlEReviewed = currentRlE.getAttributeValuesList(ReactomeJavaConstants.reviewed).size();
+        int previousRlEReviewed = previousRlE.getAttributeValuesList(ReactomeJavaConstants.reviewed).size();
+
+        // The actual QA check
+        return !sameRegulatedBy && currentRlEReviewed == previousRlEReviewed;
     }
 
     /**
--- a/src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
+++ b/src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
@@ -45,8 +45,14 @@ public class NewRegulationChecker extends AbstractQACheck implements ChecksTwoDa
         for (GKInstance currentRlE : currentRlEsWithRegulations) {
             GKInstance previousRlE = priorAdaptor.fetchInstance(currentRlE.getDBID());
             // QA check
-            if (newRegulationWithoutReviewed(currentRlE, previousRlE)){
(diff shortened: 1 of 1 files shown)


src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
```diff
@@ -0,0 +1,84 @@
...
+        QAReport report = new QAReport();
+        report.setColumnHeaders("DBID","DisplayName", "SchemaClass", "Issue", "MostRecentAuthor");
+        // Find all ReactionlikeEvents that have a filled 'regulatedBy' attribute in current slice.
+        Set<GKInstance> currentRlEsWithRegulations = getRlEsWithRegulationInstances(dba);
+
+        for (GKInstance currentRlE : currentRlEsWithRegulations) {
+            GKInstance previousRlE = priorAdaptor.fetchInstance(currentRlE.getDBID());
+            // QA check
+            if (newRegulationWithoutReviewed(currentRlE, previousRlE)){
+                report.addLine(currentRlE.getDBID().toString(), currentRlE.getDisplayName(), currentRlE.getSchemClass().getName(), "ReactionlikeEvent with new regulatedBy instance but has not yet been reviewed", QACheckerHelper.getLastModificationAuthor(currentRlE));
```
jweiser, line 49: Maybe for readability, it might be helpful to have each column value on its own line.  You could do something like:

```
report.addLine(
    currentRlE.getDBID().toString(),
    currentRlE.getDisplayName(),
    ...
);
```

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
```diff
@@ -0,0 +1,84 @@
...
+     * the number of reviewed instances is the same, it is flagged to indicate that this ReactionlikeEvent still needs to reviewed before
+     * it can be released.
+     * @param currentRlE GKInstance -- ReactionlikeEvent from current slice.
+     * @param previousRlE GKInstance -- Same ReactionlikeEvent, but from previous slice.
+     * @return boolean -- True if a new, unreviewed regulation instance has been added.
+     * @throws Exception -- DBA exceptions.
+     */
+    private boolean newRegulationWithoutReviewed(GKInstance currentRlE, GKInstance previousRlE) throws Exception {
+        // If previousRlE does not exist, we ignore it since that does not pertain to this particular QA check.
+        if (previousRlE != null) {
```
jweiser, line 67: You could consider a guard clause here to return if the `previousRlE` is null.  

```
if (previousRlE == null) {
    return false;
}
```

It would reduce the indentation of the main logic in the if block.  It's a style judgement call, though.

SolomonShorser-OICR:
I'm fine with this, but @guanmingwu / @lbeckman314  should probably have the final say.

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
cookersjs: I don't mind extra indentation, I like having fewer exit points

jweiser: I think you could keep the number of exit points the same if you were to use the guard clause and change line 74 to:

```
return currentRlERegulationCount > previousRlERegulationCount && 
currentRlEReviewedCount == previousRlEReviewedCount
```

In this case, though, the benefit is small, so I'm okay with leaving it as is.

Guard clauses are usually considered good practice, though.  This article addresses the argument of not using guard clauses due to too many exit points: https://medium.com/better-programming/refactoring-guard-clauses-2ceeaa1a9da.  The counter-argument they give is with guard clauses, all the additional exit points are at the beginning of the method and show the conditions under which the "normal" logic isn't followed.  It can also reduce indentation and conditionals within conditionals making code hard to reason about.  The second point isn't relevant in this case since it's only one level of nesting, but it shows up a lot of the time.

I try to use guard clauses when it helps reduce the code complexity and improve readability since it is a well established refactoring principle: https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
```diff
@@ -0,0 +1,84 @@
...
+            int currentRlERegulationCount = currentRlE.getAttributeValuesList(ReactomeJavaConstants.regulatedBy).size();
+            int currentRlEReviewedCount = currentRlE.getAttributeValuesList(ReactomeJavaConstants.reviewed).size();
+            int previousRlERegulationCount = previousRlE.getAttributeValuesList(ReactomeJavaConstants.regulatedBy).size();
+            int previousRlEReviewedCount = previousRlE.getAttributeValuesList(ReactomeJavaConstants.reviewed).size();
+
+            // The actual QA check
+            if (currentRlERegulationCount > previousRlERegulationCount && currentRlEReviewedCount == previousRlEReviewedCount) {
```
guanmingwu, line 74: Simply based on the numbers of regulatedBy instances may not be right: e.g. the curator may delete two regulateBy instances and then add one or just replace one and then add another one. You may have to check the actual content here. Please see this as an example: https://github.com/reactome/CuratorTool/blob/b1599c347d8761e02316d40638758344585b6b29/src/org/gk/model/InstanceUtilities.java#L569.

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
```diff
@@ -0,0 +1,84 @@
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private Set<GKInstance> getRlEsWithRegulationInstances(MySQLAdaptor dba) throws Exception {
+       return (HashSet<GKInstance>) dba.fetchInstanceByAttribute(ReactomeJavaConstants.ReactionlikeEvent, ReactomeJavaConstants.regulatedBy, "!=", "null");
```
guanmingwu, line 82: Usually I use something like this:

dba.fetchInstanceByAttribute(ReactomeJavaConstants.ReactionlikeEvent,
                                                                   ReactomeJavaConstants.entityFunctionalStatus,
                                                                   "IS NOT NULL",
                                                                   null);

I never use this query before. But I assume that you have tested it.

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
cookersjs: Yea that change makes sense, I have made it

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
cookersjs: Both seem to work the same in this case, but I have changed it to your suggestion to keep in it in line with typical use

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
cookersjs: @guanmingwu I tried to use the `compareAttValues` method (https://github.com/reactome/CuratorTool/blob/b1599c347d8761e02316d40638758344585b6b29/src/org/gk/model/InstanceUtilities.java#L626) in InstanceUtilities but it isn't working as expected.

When I try `compareAttValues(currentRlE, previousRlE, regulatedBySchemaAttribute)` I get exception 'Cannot cast GKInstance to List'. This exception makes sense based on the method. But what do I provide to the method to compare the 'regulatedBy' values of two ReactionlikeEvent instances? 

I've tried submitting both RlE's "regulatedBy" Lists to the method, and I get an unexpected exception `Invalid attribute 'regulatedBy' for class 'Reaction'`, which doesn't seem right.

I've tried submitting both RlEs as a single-entry List. This doesn't throw an Exception but it always returns true, which is wrong.

Is this method meant to be used this way? Anyways, this is a relatively simple task so I'm just going to compare the values of 'regulatedBy' within this code.

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
```diff
@@ -62,23 +68,57 @@ public QAReport executeQACheck() throws Exception
...
+
         // If previousRlE does not exist, we ignore it since that does not pertain to this particular QA check.
-        if (previousRlE != null) {
-            int currentRlERegulationCount = currentRlE.getAttributeValuesList(ReactomeJavaConstants.regulatedBy).size();
-            int currentRlEReviewedCount = currentRlE.getAttributeValuesList(ReactomeJavaConstants.reviewed).size();
-            int previousRlERegulationCount = previousRlE.getAttributeValuesList(ReactomeJavaConstants.regulatedBy).size();
-            int previousRlEReviewedCount = previousRlE.getAttributeValuesList(ReactomeJavaConstants.reviewed).size();
+        if (previousRlE == null) {
+            return false;
+        } else {
```
jweiser, line 76: Since the if condition is a guard and returns from the method, the else can actually be removed because the method execution won't go any farther when it `previousRlE` is null.

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
```diff
@@ -62,23 +68,57 @@ public QAReport executeQACheck() throws Exception
...
+     */
+    private boolean isSameRegulatedByValues(List<GKInstance> regulatedByInstancesCurrent, List<GKInstance> regulatedByInstancesPrevious) {
+        List<GKInstance> list1Copy = new ArrayList(regulatedByInstancesCurrent);
+        List<GKInstance> list2Copy = new ArrayList(regulatedByInstancesPrevious);
+        GKInstance instance1;
+        GKInstance instance2;
+        for (Iterator it = list1Copy.iterator(); it.hasNext();) {
+            instance1 = (GKInstance) it.next();
+            for (Iterator it1 = list2Copy.iterator(); it1.hasNext();) {
+                instance2 = (GKInstance) it1.next();
```
jweiser, line 106: I think you can put the declaration of `instance1` and `instance2` inline with their assignments on line 104 and 106

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
guanmingwu: no equation marks around the last parameter null, please

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
```diff
@@ -0,0 +1,120 @@
...
+        boolean sameRegulatedBy = isSameRegulatedByValues(
+                currentRlE.getAttributeValuesList(ReactomeJavaConstants.regulatedBy),
+                previousRlE.getAttributeValuesList(ReactomeJavaConstants.regulatedBy)
+        );
+
+        int currentRlEReviewed = currentRlE.getAttributeValuesList(ReactomeJavaConstants.reviewed).size();
+        int previousRlEReviewed = previousRlE.getAttributeValuesList(ReactomeJavaConstants.reviewed).size();
+
+        // The actual QA check
+        return !sameRegulatedBy && currentRlEReviewed == previousRlEReviewed;
```
guanmingwu, line 89: The count-based comparison is always error-prone. We have to assume the curator doesn't make any change to the reviewed slot. Probably it is fine for this specific slot.

src/main/java/org/reactome/release/qa/check/NewRegulationChecker.java
```diff
@@ -0,0 +1,120 @@
...
+            return false;
+        }
+        Set<Long> list1DbIds = getInstanceListDBIDs(regulatedByInstancesCurrent);
+        Set<Long> list2DbIds = getInstanceListDBIDs(regulatedByInstancesPrevious);
+        return list1DbIds.equals(list2DbIds);
+    }
+
+    // Return all unique DBIDs from instances in the list
+    private Set<Long> getInstanceListDBIDs(List<GKInstance> list) {
+        return list.stream().map(GKInstance::getDBID).collect(toSet());
```
guanmingwu, line 110: Thought the set comparison should work, however, better to use a sorted list for defensive programming.

PR #17, opened 2020-05-22
Feature/dev 1837 and 1838

cookersjs:
Basically the same PR as https://github.com/reactome/CuratorTool/pull/29, with a few minor changes.

The output report format has been changed from `.txt` to `.tsv` -- Curation had pointed out that some reports were not uploading to GDrive properly and this may have been why.

All new QA checks have been annotated to `@SliceQATests.`


cookersjs:
@jweiser I've addressed the requested changes that you commented in the original PR (https://github.com/reactome/CuratorTool/pull/29). For the most part I made your suggested changes, though I decided to go against the `triple-for-loop` helper method you suggested -- I gave my reasoning in that thread.

src/main/java/org/reactome/release/qa/check/HumanReactionsWithNonHumanComplexesWithHumanComponentsCheck.java
```diff
@@ -0,0 +1,100 @@
...
+ * QA check that finds non-human Complexes that contain human Components, that are members of human, non-manually inferred ReactionlikeEvents.
+ */
+
+@SliceQATest
+public class HumanReactionsWithNonHumanComplexesWithHumanComponentsCheck extends AbstractQACheck {
+    @Override
+    public QAReport executeQACheck() throws Exception {
+        QAReport report = new QAReport();
+        QACheckerHelper.setHumanSpeciesInst(dba);
+        QACheckerHelper.setSkipList(null);
```
jweiser, line 25: I'm just wondering if this line, in this and other classes, is needed since the skip list will be an empty list by default?

src/main/java/org/reactome/release/qa/check/HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntities.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Set;
+
+/**
+ * Flags human ReactionlikeEvents that do not have a populated disease attribute, but have non-human participants.
+ */
+
+@SliceQATest
+public class HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntities extends AbstractQACheck {
+
+    private static final String innateImmunityPathwayDbId = "168249";
```
jweiser, line 22: Nice, thanks for adding the named constant.  For my famous PR nit-picking, could the name be upper case with underscores instead of camel case (i.e. `INNATE_IMMUNITY_PATHWAY_DB_ID`)?

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -24,6 +24,8 @@
     
 	public static final String IS_NOT_NULL = "IS NOT NULL";
 	public static final String IS_NULL = "IS NULL";
+	private static List<String> skiplistDbIds = new ArrayList<>();
```
jweiser, line 27: I was wondering if the skip list db id list might be better as a member of the individual checks rather than in the QACheckerHelper?  Having it as a static member means it will be a single object shared among the checks.  This shouldn't be a problem if the checks are run serially and the skip lists are set at each check, but it might be problematic if the checks are run in parallel.  It might make more sense for each check to own its own skip list.  Maybe it could be a member in the common parent class of `AbstractQACheck`?

src/main/java/org/reactome/release/qa/check/HumanReactionsWithNonHumanComplexesWithHumanComponentsCheck.java
cookersjs: Good catch, I'll remove that line

src/main/java/org/reactome/release/qa/check/HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntities.java
cookersjs: Another good catch -- I forgot about that for global variables

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
SolomonShorser-OICR: That would make sense if different QA checks could have different skip lists.

src/main/java/org/reactome/release/qa/check/HumanReactionsWithNonHumanComplexesWithHumanComponentsCheck.java
```diff
@@ -7,24 +7,24 @@
...
 
 /**
  * QA check that finds non-human Complexes that contain human Components, that are members of human, non-manually inferred ReactionlikeEvents.
  */
 
 @SliceQATest
 public class HumanReactionsWithNonHumanComplexesWithHumanComponentsCheck extends AbstractQACheck {
+
+    // Rather than overload multiple methods that use/don't use a skiplist, an empty List is used here.
+    private static List<String> EMPTY_SKIP_LIST = new ArrayList<>();
```
jweiser, line 20: Minor thing, but since this is in a class meant to be used as an instance now instead of a utility static method based class, the `static` qualifier could be removed.

src/main/java/org/reactome/release/qa/check/HumanReactionsWithNonHumanPhysicalEntitiesWithoutDiseaseCheck.java
```diff
@@ -7,21 +7,26 @@
...
+
 /**
  * Flags all human Reactions that contain non-disease PhysicalEntities that are non-human, or are human PhysicalEntities with relatedSpecies.
  */
 
 @SliceQATest
 public class HumanReactionsWithNonHumanPhysicalEntitiesWithoutDiseaseCheck extends AbstractQACheck {
 
+    // Rather than overload multiple methods that use/don't use a skiplist, an empty List is used here.
+    private static List<String> EMPTY_SKIP_LIST = new ArrayList<>();
```
jweiser, line 21: Same here with removing `static`

src/main/java/org/reactome/release/qa/check/HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntities.java
```diff
@@ -19,18 +16,19 @@
 @SliceQATest
 public class HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntities extends AbstractQACheck {
 
-    private static final String innateImmunityPathwayDbId = "168249";
+    private static final String INNATE_IMMUNITY_PATHWAY_DBID = "168249";
+    private static List<String> skiplistDbIds = new ArrayList<>();
```
jweiser, line 20: In this case, `static` should definitely be removed since `skiplistDbIds` will be receiving values in `executeQAChceck`.  Every instance of this check should have its own skip list and rather than shared using `static`.

src/main/java/org/reactome/release/qa/check/HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntities.java
```diff
@@ -19,18 +16,19 @@
...
 
     @Override
     public QAReport executeQACheck() throws Exception {
         QAReport report = new QAReport();
-        QACheckerHelper.setSkipList(Arrays.asList(innateImmunityPathwayDbId)); // Innate Immunity Pathway
         QACheckerHelper.setHumanSpeciesInst(dba);
+        skiplistDbIds.add(INNATE_IMMUNITY_PATHWAY_DBID);
```
jweiser, line 26: You could make this `this.skiplistDbIds` so it is clear it is an instance variable.

src/main/java/org/reactome/release/qa/check/NonHumanEventsNotManuallyInferredChecker.java
```diff
@@ -7,21 +7,26 @@
...
+import java.util.List;
+
 /**
  *  Finds all non-human Events that are not used for manual inference (ie. inferredFrom referral is null)
  */
 
 @SliceQATest
 public class NonHumanEventsNotManuallyInferredChecker extends AbstractQACheck {
 
+    private static List<String> skiplistDbIds = new ArrayList<>();
```
jweiser, line 20: Same as above re: removing `static`

src/main/java/org/reactome/release/qa/check/NonHumanEventsNotManuallyInferredChecker.java
```diff
@@ -7,21 +7,26 @@
...
+
     @Override
     public QAReport executeQACheck() throws Exception {
         QAReport report = new QAReport();
         QACheckerHelper.setHumanSpeciesInst(dba);
-        QACheckerHelper.setSkipList(QACheckerHelper.getNonHumanPathwaySkipList());
+        skiplistDbIds = QACheckerHelper.getNonHumanPathwaySkipList();
```
jweiser, line 26: Same as above re: adding `this.`

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -564,6 +558,6 @@ public static void setHumanSpeciesInst(MySQLAdaptor dba) throws Exception {
 	 */
 	public static String getInstanceAttributeNameForOutputReport(GKInstance instance, String attribute) throws Exception {
 		GKInstance attributeInstance = (GKInstance) instance.getAttributeValue(attribute);
-		return attributeInstance != null ? attributeInstance.getDisplayName() : null;
+		return attributeInstance != null ? attributeInstance.getDisplayName() : "N/A";
```
jweiser, line 561: Excellent :)

src/main/java/org/reactome/release/qa/check/NonHumanReactionsWithHumanPhysicalEntitiesCheck.java
```diff
@@ -18,16 +16,18 @@
...
 public class NonHumanReactionsWithHumanPhysicalEntitiesCheck extends AbstractQACheck {
 
+    private static List<String> skiplistDbIds = new ArrayList<>();
+
     @Override
     public QAReport executeQACheck() throws Exception {
         QAReport report = new QAReport();
         QACheckerHelper.setHumanSpeciesInst(dba);
-        QACheckerHelper.setSkipList(QACheckerHelper.getNonHumanPathwaySkipList());
+        skiplistDbIds = QACheckerHelper.getNonHumanPathwaySkipList();
```
jweiser, line 25: Same as above re: `this.`

src/main/java/org/reactome/release/qa/check/NonHumanReactionsWithHumanPhysicalEntitiesCheck.java
```diff
@@ -18,16 +16,18 @@
 @SliceQATest
 public class NonHumanReactionsWithHumanPhysicalEntitiesCheck extends AbstractQACheck {
 
+    private static List<String> skiplistDbIds = new ArrayList<>();
```
jweiser, line 19: Same as above re: removing `static`

src/main/java/org/reactome/release/qa/check/HumanReactionsWithNonHumanPhysicalEntitiesWithoutDiseaseCheck.java
SolomonShorser-OICR: If `EMPTY_SKIP_LIST` is really going to be used instead of a few simple overloads, then it probably should not be duplicated in `HumanReactionsWithNonHumanComplexesWithHumanComponentsCheck` as well as `HumanReactionsWithNonHumanPhysicalEntitiesWithoutDiseaseCheck`. Both classes extend `AbstractQACheck` so `EMPTY_SKIP_LIST` should be moved to that, or if it is not appropriate for _all_ `AbstractQACheck`s to have access to `EMPTY_SKIP_LIST`, then a new class or interface could be introduced, say, `HasNoSkipList` and then `HasNoSkipList` could extend `AbstractQACheck` and then `HumanReactionsWithNonHuman...Check` classes could extend `HasNoSkipList`.

jweiser: @SolomonShorser-OICR I was thinking that, too.  I like that approach but maybe it's best to wait for the next PR?  I get the sense there are existing checks outside the scope of this PR that will benefit from that kind of class or interface structure as well.

cookersjs: I moved EMPTY_SKIP_LIST to `AbstractQACheck`

src/main/java/org/reactome/release/qa/common/AbstractQACheck.java
```diff
@@ -9,6 +11,7 @@
     private final static Logger logger = Logger.getLogger(AbstractQACheck.class);
 
     protected MySQLAdaptor dba;
+    protected List<String> EMPTY_SKIP_LIST = new ArrayList<>();
```
jweiser, line 14: Looks good!  Since this is meant to be used as a constant, you could have it be `final` and optionally `public`.  You could also set it to an unmodifiable list using `Collections.emptyList()`: https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#emptyList--

cookersjs: I did all except make it public -- to keep it in line with how dba is initialized

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -24,6 +24,7 @@
     
 	public static final String IS_NOT_NULL = "IS NOT NULL";
 	public static final String IS_NULL = "IS NULL";
+	private static GKInstance humanSpeciesInst = new GKInstance();
```
jweiser, line 27: Instead of having a human species instance set to an empty `GKInstance`, it might be better to remove this variable and its setter method and have a getter method instead that takes a `dba` object and returns a new human species instance based on the `dba` object passed in: `dba.fetchInstance(48887L)`.  You could cache the value, but it is probably better to just return a new species instance since it could vary by the `dba` passed in and to cache by `dba` object seems like overkill.

src/main/java/org/reactome/release/qa/common/AbstractQACheck.java
SolomonShorser-OICR: Hmmm... using `Collections.emptyList()` might be a good idea, if you want to have a constant be an empty list.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
cookersjs: So I've implemented this (there was already a 'getHuman' method actually in QACheckerHelper), but it really causes the tests to slow down. I assume its because it queries the DB for almost every single instance that it is checking. It used to take 1-2 minutes to run all 5 tests, now its taking 8-10 minutes to run them. Are we OK with such a tradeoff in runtime to avoid caching the Human instance?

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -54,14 +54,7 @@ public static boolean isChimeric(GKInstance rle) throws Exception {
...
 	public static GKInstance getHuman(MySQLAdaptor dba) throws Exception {
-	    Collection<GKInstance> c = dba.fetchInstanceByAttribute(ReactomeJavaConstants.Species,
-	            ReactomeJavaConstants._displayName,
-	            "=", 
-	            "Homo sapiens");
-	    if (c == null || c.size() == 0)
-	        throw new IllegalStateException("Cannot find species Homo sapiens in the database, " + 
-	                                        dba.getDBName() + "@" + dba.getDBHost());
-	    return c.iterator().next();
+	    return dba.fetchInstance(HUMAN_SPECIES_INSTANCE_DBID);
```
SolomonShorser-OICR, line 57: I would recommend a private static GKInstance variable in this class, call it `humanInst`. When `getHuman` is called, check to see if `humanInst` is null, and if it is, then do the `dba.fetchInstance` and then _set_ the private static `humanInst` to whatever comes back from the database. Once it's been set, you can just return the same object every time - no need to keep sending the same query to the database every time a "human" is needed.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
SolomonShorser-OICR: No, I don't think that's a good trade-off. There is no need to keep querying for the same object again and again and again and again.

SolomonShorser-OICR: Also: what was the original problem with a cached object for a single species?

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -54,14 +53,14 @@ public static boolean isChimeric(GKInstance rle) throws Exception {
...
-	            "=", 
-	            "Homo sapiens");
-	    if (c == null || c.size() == 0)
-	        throw new IllegalStateException("Cannot find species Homo sapiens in the database, " + 
-	                                        dba.getDBName() + "@" + dba.getDBHost());
-	    return c.iterator().next();
+		Collection<GKInstance> c = dba.fetchInstanceByAttribute(ReactomeJavaConstants.Species,
+				ReactomeJavaConstants._displayName,
+				"=",
+				"Homo sapiens");
```
SolomonShorser-OICR, line 59: Could the result of this query be cached somewhere so that it doesn't need to be executed every time the `getHuman` method is called?

cookersjs: Actually this is a method that was created from Fred, I just reverted it after coming up with a solution that didn't require checking of a GKInstance method. That being said, I can implement that caching suggestion since the method is terribly inefficient

jweiser: @cookersjs Actually, it doesn't look like your code is even calling this method anymore?

cookersjs: No its not. The story: This is a method that had already existed in release-qa/.../QACheckerHelper, I just hadn't realized it. During the earlier iterations of work today where the code still had places checking the Human instance, I decided to utilize this code instead of create new (and redundant), Human-instance-getting code. The only problem was that the method was really slow due to the `fetchSpeciesByAttribute` usage -- so I rewrote the method to search specifically for the instance with DbId 48887. This was quicker and got the runtime down to the usual 1-2 minutes. 

After the dust settled and we came up with a way to check for Homo sapiens without using the Homo sapiens instance, I figured it might as revert the code back to the state it was, since it wasn't mine and I wasn't using it anymore.

jweiser: Excellent story :)

SolomonShorser-OICR: If this original code is still used by other QAChecks, I'd leave it as it is, but with a TODO note for future development, maybe saying "this could be much more efficient if the GKInstance were only fetched once and cached".

This PR is (mostly) for _your_ code, and if it's no longer necessary to change it, then don't change it.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -52,16 +51,17 @@ public static boolean isChimeric(GKInstance rle) throws Exception {
         return true;
     }
 	
+	// TODO: If this method is being heavily used by a QA check, consider using 'fetchInstance(48887L)`. It's much quicker than a global DB search.
```
SolomonShorser-OICR, line 54: This is an important comment, and since it's at the method-level, could you please upgrade it to Javadoc? I think my "TODO" idea was about leaving a note to implement a cached instance so that there is only ever one database query.

SolomonShorser-OICR:
I have no further comment, I'll leave this to @guanmingwu and @lbeckman314 to approve for merging.

src/main/java/org/reactome/release/qa/Main.java
```diff
@@ -99,8 +99,8 @@ public static void main(String[] args) throws Exception {
             }
             else {
                 String fileName = check.getDisplayName();
-                qaReport.output(fileName + ".txt", output.getAbsolutePath());
-                logger.info("Check "+ output.getAbsolutePath() + "/" + fileName + ".txt for report details.");
+                qaReport.output(fileName + ".tsv", output.getAbsolutePath());
```
guanmingwu, line 102: Be aware: changing file extension may break html links generated in the weekly report. Please make sure.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -225,5 +232,339 @@ public static String getAttributeTableName(String clsName,
...
+	 * @return Set<GKInstance> -- All distinct PhysicalEntitys in ReactionlikeEvent
+	 * @throws Exception -- Thrown by MySQLAdaptor
+	 */
+	public static Set<GKInstance> findAllPhysicalEntitiesInReaction(GKInstance reaction) throws Exception {
+		Set<GKInstance> reactionPEs = new HashSet<>();
+		reactionPEs.addAll(findAllInputAndOutputPEs(reaction));
+		reactionPEs.addAll(findAllCatalystPEs(reaction));
+		reactionPEs.addAll(findAllRegulationPEs(reaction));
+		return reactionPEs;
+	}
```
guanmingwu, line 287: Please use org.gk.model.InstanceUtilities.getReactionparticipants(GKInstance) so that we will have the same logic across all codebase.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -225,5 +232,339 @@ public static String getAttributeTableName(String clsName,
...
+
+	/**
+	 * Finds all PhysicalEntities that exist in a ReactionlikeEvent's inputs and outputs.
+	 * @param reaction GKInstance -- ReactionlikeEvent that is being searched for PhysicalEntities.
+	 * @return Set<GKInstance> -- All distinct PhysicalEntitys in ReactionlikeEvent's inputs and outputs.
+	 * @throws Exception -- Thrown by MySQLAdaptor
+	 */
+	public static Set<GKInstance> findAllInputAndOutputPEs(GKInstance reaction) throws Exception {
```
guanmingwu, line 295: Should the method name changed since you have a constraint in your input/output selection? Also if this static method is used by a specific class, could it be merged into that class instead here as a generic method to be used everywhere?

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -225,5 +232,339 @@ public static String getAttributeTableName(String clsName,
...
+				for (GKInstance attributePE : (Collection<GKInstance>) catalyst.getAttributeValuesList(attribute)) {
+					// QA checks calling these methods are concerned with PhysicalEntities that have a species attribute.
+					if (hasSpeciesAttribute(attributePE)) {
+						catalystPEs.add(attributePE);
+						catalystPEs.addAll(findAllPhysicalEntities(attributePE));
+					}
+				}
+			}
+		}
+		return catalystPEs;
```
guanmingwu, line 331: See the comment for the above method.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -225,5 +232,339 @@ public static String getAttributeTableName(String clsName,
...
+				for (GKInstance attributePE : (Collection<GKInstance>) regulation.getAttributeValuesList(attribute)) {
+					// QA checks calling these methods are concerned with PhysicalEntities that have a species attribute.
+					if (hasSpeciesAttribute(attributePE)) {
+						regulationPEs.add(attributePE);
+						regulationPEs.addAll(findAllPhysicalEntities(attributePE));
+					}
+				}
+			}
+		}
+		return regulationPEs;
```
guanmingwu, line 356: Same here as the above method.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -225,5 +232,339 @@ public static String getAttributeTableName(String clsName,
...
+	}
+
+	/**
+	 * Finds all PhysicalEntities that exist within an EntitySet by searching for all member PEs (via 'hasMember' attribute)
+	 * and all candidate PEs (if entitySet is a CandidateSet; via 'hasCandidate attribute).
+	 * @param entitySet GKInstance -- EntitySet instance that will be searched for all PhysicalEntities.
+	 * @return Set<GKInstance> -- All distinct PhysicalEntities that are found in EntitySet.
+	 * @throws Exception -- Thrown by MySQLAdaptor.
+	 */
+	public static Set<GKInstance> findEntitySetPhysicalEntities(GKInstance entitySet) throws Exception {
```
guanmingwu, line 437: Please consider using static methods in InstanceUtilities to avoid any duplication.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -225,5 +232,339 @@ public static String getAttributeTableName(String clsName,
...
+	}
+
+	/**
+	 * Checks if incoming DatabaseObject (Event or PhysicalEntity) has single Homo sapiens species.
+	 * @param databaseObject GKInstance -- Event or PhysicalEntity instance being checked for only Homo sapiens species.
+	 * @return boolean -- true if databaseObject only has a single, Homo sapiens species instance, false if not.
+	 * @throws Exception -- Thrown by MySQLAdaptor.
+	 */
+	public static boolean isHumanDatabaseObject(GKInstance databaseObject) throws Exception {
+		List<GKInstance> objectSpecies = databaseObject.getAttributeValuesList(ReactomeJavaConstants.species);
```
guanmingwu, line 471: Since this is a generic method, please check if the species is a valid attribute first.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -225,5 +232,339 @@ public static String getAttributeTableName(String clsName,
...
+	}
+
+	/**
+	 * Checks if the incoming databaseObject has a popalated disease attribute.
+	 * @param databaseObject GKInstance -- Instance to be checked for populated disease attribute.
+	 * @return boolean -- true if has filled disease attribute, false if not.
+	 * @throws Exception -- Thrown by MySQLAdaptor.
+	 */
+	public static boolean hasDisease(GKInstance databaseObject) throws Exception {
+		return databaseObject.getAttributeValue(ReactomeJavaConstants.disease) != null;
```
guanmingwu, line 514: Again check if the disease is a valid attribute.

src/main/java/org/reactome/release/qa/Main.java
guanmingwu: Here it is an example: https://curator.reactome.org/QAReports/20200525/ReleaseQA/Attribute_Has_Only_One_Value_wug.html. The html file was generated based on these output files.

cookersjs: I'll revert it, my mistake

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
cookersjs: OK, I'll do that. Does this method get all catalyst and regulation participants as well?

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
cookersjs: Each of these 'findPhysicalEntities' methods are used by each of the new QA checks, so they do need to be accessed generically. If I am going to switch to using InstanceUtilities methods though, the amount of code in QACheckerHelper will be reduced quite a bit

cookersjs:
@guanmingwu @SolomonShorser-OICR @jweiser -- I've reverted the report from tsv back to txt, and I've converted the methods to use InstanceUtilities to get all participants. Only thing I added in addition was to get the participants for catalyst/regulation activeUnits to address the things found from my code.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -252,6 +247,16 @@ public static String getAttributeTableName(String clsName,
...
+	/**
+	 * Checks if incoming Event is manually inferred by checking inferredFrom referral.
+	 * @param event GKInstance -- Event instance being checked for inferredFrom referral.
+	 * @return boolean -- true if inferredFrom referral exists, false if not.
+	 * @throws Exception -- Thrown by MySQLAdaptor.
+	 */
+	private static boolean manuallyInferred(GKInstance event) throws Exception {
+		return event.getReferers(ReactomeJavaConstants.inferredFrom) != null;
+	}
+
```
jweiser, line 259: Maybe you could add a note that this check is only valid for instances coming from a curation database (i.e. gk_central or a slice) rather than a release database that uses `inferredFrom` for electronic inference.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -273,192 +278,43 @@ public static String getAttributeTableName(String clsName,
...
-		if (multiPEInstance.getSchemClass().isa(ReactomeJavaConstants.Complex)) {
-			constituentPEs.addAll(multiPEInstance.getAttributeValuesList(ReactomeJavaConstants.hasComponent));
-		} else if (multiPEInstance.getSchemClass().isa(ReactomeJavaConstants.Polymer)) {
-			constituentPEs.addAll(multiPEInstance.getAttributeValuesList(ReactomeJavaConstants.repeatedUnit));
-		}
-		return constituentPEs;
-	}
+	public static Set<GKInstance> getAllReactionParticipantsIncludingCatalystAndRegulations(GKInstance reaction) throws Exception {
+		Set<GKInstance> reactionPEs = new HashSet<>();
+		reactionPEs.addAll(InstanceUtilities.getReactionParticipants(reaction));
```
jweiser, line 291: Maybe this method could get renamed to something like `getReactionInputsAndOutputs`.  The name could be confusing or misleading if this doesn't include catalysts and regulators of the reaction.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -273,192 +278,43 @@ public static String getAttributeTableName(String clsName,
...
-	public static boolean manuallyInferred(GKInstance event) throws Exception {
-		return event.getReferers(ReactomeJavaConstants.inferredFrom) != null;
+	public static Set<GKInstance> getPhysicalEntityContainedInstances(GKInstance reactionPE) throws Exception {
+		return InstanceUtilities.getContainedInstances(
+				reactionPE,
+				ReactomeJavaConstants.hasMember,
+				ReactomeJavaConstants.hasCandidate,
+				ReactomeJavaConstants.hasComponent,
+				ReactomeJavaConstants.repeatedUnit,
+				ReactomeJavaConstants.activeUnit);
```
jweiser, line 317: Nice - I like that this is one line per attribute!  One nit-picky style thing you could consider is putting the closing parenthesis on its own line with the same indentation as line 311.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -468,15 +324,17 @@ public static boolean manuallyInferred(GKInstance event) throws Exception {
...
+				databaseObject.getAttributeValuesList(ReactomeJavaConstants.species) : Collections.emptyList();
 		return objectSpecies.size() == 1 && hasHumanSpecies(objectSpecies);
 	}
 
 	/**
 	 * Checks if the incoming DatabaseObject (Event or PhysicalEntity) is non-human.
 	 * @param databaseObject GKInstance -- Event or PhysicalEntity to be checked for non-human species attribute.
 	 * @return boolean -- true if instance has species, and if none of the species are human, false if not.
-	 * @throws Exception
+	 * @throws Exception -- Thrown by MysqlAdaptor.
```
jweiser, line 337: Nit-pick du jour: Could "sql" be captialized for consistency?

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -273,192 +278,43 @@ public static String getAttributeTableName(String clsName,
...
-	}
-
-	/**
-	 * Helper method that returns either a Complexes' components or Polymers' repeatedUnits.
-	 * @param multiPEInstance GKInstance -- Either a Complex or Polymer instance.
-	 * @return Set<GKInstance> -- All components or repeatedUnits from the incoming instance.
+	 * This method utilizes the InstanceUtilities method for finding all reaction participants. It also checks for the
+	 * 'activeUnit' participant in catalysts/regulations that can be flagged by certain QA checks. Generally speaking,
+	 * the 'activeUnit' instance should be found in the 'physicalEntity' of catalysts/regulations. Sometimes this is not
+	 * the case, when the wrong activeUnit is mistakenly checked in. This will be flagged by QA tests that call this method.
```
jweiser, line 284: Just curious - If this method returns all participants, how is this method used by other methods to determine if the activeUnit specifically is wrong?

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
cookersjs: That doesn't quite capture what this method is trying to do, maybe I'll rename it to be 'IncludingActiveUnit' instead

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
cookersjs: As we discussed yesterday (but I'm posting here for posterity) -- the method isn't meant to specifically target activeUnits and check that they match with their catalyst/regulation physicalEntity, but rather the nature of some of the QA tests that call this method is to flag PhysicalEntity instance types (such as would be found in the activeUnit attribute) that have a differing species from a Reaction they are associated with.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
cookersjs: Changed it to say 'IncludingActiveUnits'

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -225,5 +227,206 @@ public static String getAttributeTableName(String clsName,
...
+	 * @param reaction GKInstance -- ReactionlikeEvent that will be checked for all participants.
+	 * @return Set<GKInstance> -- All distinct participants in the incoming ReactionlikeEvent.
+	 * @throws Exception -- Thrown by MySQLAdaptor.
+	 */
+	public static Set<GKInstance> getAllReactionParticipantsIncludingActiveUnits(GKInstance reaction) throws Exception {
+		Set<GKInstance> reactionPEs = new HashSet<>();
+		reactionPEs.addAll(InstanceUtilities.getReactionParticipants(reaction));
+		// Catalysts/Regulations also added to check for incorrect activeUnits.
+		reactionPEs.addAll(reaction.getAttributeValuesList(ReactomeJavaConstants.catalystActivity));
+		reactionPEs.addAll(reaction.getAttributeValuesList(ReactomeJavaConstants.regulatedBy));
```
jweiser, line 297: So, my understanding is the lines getting the catalyst activities and regulations are so they can be investigated to grab the physical entities in the `activeUnit` attribute not grabbed by the call to `InstanceUtilities.getReactionParticipants`?

If this is right, I find it a bit confusing to add the catalyst activity and regulation instances directly to the `reactionPEs` set since they are not Physical Entity instances themselves but contain physical entities.  Would it be possible to hold them in a different set that is sent a method responsible only for querying activeUnit for physical entities and returning them to be added to the `reactionPEs`?  Then you could send the elements of `reactionPEs` to be recursed to get physical entities contained physical entities.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -225,5 +227,206 @@ public static String getAttributeTableName(String clsName,
...
+	 */
+	public static Set<GKInstance> getPhysicalEntityContainedInstances(GKInstance reactionPE) throws Exception {
+		return InstanceUtilities.getContainedInstances(
+				reactionPE,
+				ReactomeJavaConstants.hasMember,
+				ReactomeJavaConstants.hasCandidate,
+				ReactomeJavaConstants.hasComponent,
+				ReactomeJavaConstants.repeatedUnit,
+				ReactomeJavaConstants.activeUnit
+		);
```
jweiser, line 321: Querying activeUnit seems out of place here as all other attributes are part of Physical Entity sub-classes where activeUnit is part of CatalystActivity/Regulation.  If this can be queried in a separate method, that would be good because as it is, you have non-PE instances as part of the `reactionPE` set parameter.

PR #19, opened 2020-06-22
Feature/dev 1828

lbeckman314:


lbeckman314:
Addresses [DEV-1828](https://reactome.atlassian.net/browse/DEV-1822) ('WeeklyQA check is needed to detect person instances that share the same ORCID id (crossreference).')

PR #26, opened 2020-11-06
Feature/cov2 qa checks

cookersjs:
PR containing new QA checks that were specific to the CoV-1-to-CoV-2 projection process. These QA checks are already deployed as part of weekly QA, but at the time there was a scramble going on and I hadn't refactored the code properly, so I'd left this code on the backburner. Now that we are starting the QA consolidation work, this will need to be merged.

This PR adds 5 new QA tests, all specific to CoV-1-to-CoV-2 projections. Some of these QA tests might not be needed anymore, but we will want Curation to confirm before we actually remove them from weekly QA. At any rate, we will want them to be included in source control, so we would probably be better off just having a 'remove from release-qa' branch, if that was what was preferred.


src/main/java/org/reactome/release/qa/check/CoV2EntityCheck.java
```diff
@@ -0,0 +1,94 @@
...
+                // Check if any species instance DbIds match the CoV-1 species DbId
+                if (speciesDbIds.contains(QACheckerHelper.getCoV1SpeciesDbId())) {
+                    report.addLine(getReportLine(cov2Event, inst, "CoV-1 species found in CoV-2 instance"));
+                }
+                // Check displayName as well
+                if (inst.getDisplayName().contains("V-1")
+                        || inst.getDisplayName().contains("v-1")
+                        || inst.getDisplayName().contains("v1")
+                        || inst.getDisplayName().contains("V1")) {
+                    report.addLine(getReportLine(cov2Event, inst, "CoV-1 displayName found in CoV-2 instance"));
```
SolomonShorser-OICR, line 66: So really, this checks that the instances have a name that matches the regex `[vV]-?1`? Could that be put in the description? The description at the top was not very clear.

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -0,0 +1,176 @@
...
+     * Checks if Event has a 2020 literatureReference. This is because CoV-2 evidence was very recent when this was being evaluated.
+     * @param cov2Event - GKInstance, Event contained within 'SARS-CoV-2 Infection' Pathway.
+     * @return - boolean, true if at least 1 2020 literatureReference, false if not.
+     * @throws Exception, thrown by MySQLAdaptor.
+     */
+    private boolean hasRecentLiteratureReference(GKInstance cov2Event) throws Exception {
+
+        boolean has2020LiteratureReference = false;
+        // Iterature through all literatureReference instances, checking the 'year' attribute.
+        for (GKInstance literatureReference : (Collection<GKInstance>) cov2Event.getAttributeValuesList(ReactomeJavaConstants.literatureReference)) {
```
SolomonShorser-OICR, line 111: This could be rewritten as a while loop that breaks as soon as it finds a literature reference from 2020, as there's no need to examine ALL literature references if you only care about one being from 2020.

src/main/java/org/reactome/release/qa/check/CoVEntityDiseaseCheck.java
```diff
@@ -0,0 +1,88 @@
...
+public class CoVEntityDiseaseCheck extends AbstractQACheck {
+
+    @Override
+    public QAReport executeQACheck() throws Exception {
+        QAReport report = new QAReport();
+
+        // Fetches all Events and PhysicalEntities in DB
+        Collection<GKInstance> eventsAndPhysicalEntities = new ArrayList<>();
+        eventsAndPhysicalEntities.addAll(dba.fetchInstancesByClass(ReactomeJavaConstants.Event));
+        eventsAndPhysicalEntities.addAll(dba.fetchInstancesByClass(ReactomeJavaConstants.PhysicalEntity));
```
SolomonShorser-OICR, line 36: Is it necessary to check ALL Events and PEs? Why not filter for Events and PEs with species=CoV-1 or species=Cov-2 and _then_ loop over those, checking for the disease attribute?

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
cookersjs: Fair point

src/main/java/org/reactome/release/qa/check/CoVEntityDiseaseCheck.java
cookersjs: I think it came down to it being a bit complicated to search for: You want to search `species` and `relatedSpecies` attributes, which can both be multi-value, with 'instance' values. I think I've found that searching by attributes where the value would be an instance wasn't really working using the java API

SolomonShorser-OICR: Ah, I didn't realise that both species attributes are multi-valued. I think there _is_ a way to do that with the Reactome API, but it's a bit tricky. Probably possible to write a query in SQL that returns just a list of DB_IDs and then iterate on that to fetch by DB_ID and then perform the other checks... I usually like to leave such things as filtering to the database, because it's _good_ at that (sort of what it was made for), but if this already works and isn't too slow and the API makes it too hard to do it, then you might as well leave it... ;)

src/main/java/org/reactome/release/qa/check/CoV2EntityCheck.java
```diff
@@ -0,0 +1,94 @@
...
+
+@SliceQACheck
+public class CoV2EntityCheck extends AbstractQACheck {
+
+    @Override
+    public QAReport executeQACheck() throws Exception {
+        QAReport report = new QAReport();
+
+        // Get all contained events within 'SARS-CoV-2 Infection' Pathway, including 'SARS-CoV-2 Infection'.
+        GKInstance cov2InfectionInst = dba.fetchInstance(QACheckerHelper.getCoV2InfectionPathwayDbId());
```
jweiser, line 32: Maybe this could be `cov2InfectionPathway`?  At first glance, I thought it was the COV-2 species instance.

src/main/java/org/reactome/release/qa/check/CoV2EntityCheck.java
```diff
@@ -0,0 +1,94 @@
...
+        Collection<GKInstance> cov2Events = InstanceUtilities.getContainedEvents(cov2InfectionInst);
+        cov2Events.add(cov2InfectionInst);
+
+        for (GKInstance cov2Event : cov2Events) {
+            Set<GKInstance> cov2Instances = new HashSet<>();
```
jweiser, line 37: For context of the instances that are in the set, could this be named something like `cov2EventAndDirectParticipants`?

src/main/java/org/reactome/release/qa/check/CoV2EntityCheck.java
jweiser: This could also be written as `inst.getDisplayName().matches("^.*?[vv]-?1.*?")` (you may want to test that I got the regex correct!) to match it in a single check or for clarity you can assign the regex in a named variable liked `Cov1Regex`.

src/main/java/org/reactome/release/qa/check/CoV2EntityCheck.java
```diff
@@ -0,0 +1,94 @@
...
+                if (inst.getSchemClass().isValidAttribute(ReactomeJavaConstants.species)) {
+                    for (GKInstance speciesInst : (Collection<GKInstance>) inst.getAttributeValuesList(ReactomeJavaConstants.species)) {
+                        speciesDbIds.add(speciesInst.getDBID());
+                    }
+                }
+                if (inst.getSchemClass().isValidAttribute(ReactomeJavaConstants.relatedSpecies)) {
+                    for (GKInstance relatedSpeciesInst : (Collection<GKInstance>) inst.getAttributeValuesList(ReactomeJavaConstants.relatedSpecies)) {
+                        speciesDbIds.add(relatedSpeciesInst.getDBID());
+                    }
+                }
```
jweiser, line 56: This can be moved to its own method called something like `getSpeciesAndRelatedSpeciesDBIds(GKInstance)`

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -0,0 +1,176 @@
...
+ */
+
+@SliceQACheck
+public class CoV2InfectionPathwayEventCheck extends AbstractQACheck {
+
+    @Override
+    public QAReport executeQACheck() throws Exception {
+        QAReport report = new QAReport();
+
+        GKInstance cov2InfectionInst = dba.fetchInstance(QACheckerHelper.getCoV2InfectionPathwayDbId());
```
jweiser, line 39: Same as https://github.com/reactome/release-qa/pull/26/files#r520039300

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -0,0 +1,176 @@
...
+    private boolean hasRecentlyModifiedSummation(GKInstance cov2Event) throws Exception {
+       for (GKInstance summation : (Collection<GKInstance>) cov2Event.getAttributeValuesList(ReactomeJavaConstants.summation)) {
+            GKInstance createdInst = (GKInstance) summation.getAttributeValue(ReactomeJavaConstants.created);
+            List<GKInstance> modifieds = summation.getAttributeValuesList(ReactomeJavaConstants.modified);
+            if (modifieds.size() > 0) {
+                // At time of writing (August 2020), nothing is reported from this clause, but it is a good check.
+                GKInstance mostRecentModifiedInst = modifieds.get(modifieds.size() - 1);
+                // Only check Year/Month/Day since the changes took place more than a day before Curators got to look at them in gk_central.
+                int createdDateTime = Integer.valueOf(createdInst.getAttributeValue(ReactomeJavaConstants.dateTime).toString().split(" ")[0].replaceAll("-", ""));
+                int modifiedDateTime = Integer.valueOf(mostRecentModifiedInst.getAttributeValue(ReactomeJavaConstants.dateTime).toString().split(" ")[0].replaceAll("-", ""));
```
jweiser, line 89: This is a bit of a mouthful.  Could this be a method like `getDate(GKInstance instanceEdit)` (maybe something for release-common-lib, actually).  

Also, it's a bit unusual to have a date represented as a raw `int`.  I won't insist on changing it here, but you could use the Java built-in `LocalDate`: https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html.

It has methods for comparing dates like `isAfter(otherDate)` and `isBefore(otherDate)` and returning a boolean.

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
jweiser: I know I'm in the minority on this, but you could also consider streaming the literature references and using the method `anyMatch` to check for the condition you're looking for.

src/main/java/org/reactome/release/qa/check/InferredEvidenceTypeCheck.java
```diff
@@ -0,0 +1,59 @@
...
+        Collection<GKInstance> events = dba.fetchInstancesByClass(ReactomeJavaConstants.Event);
+        for (GKInstance event : events) {
+            // Orthoinference adds a typical EvidenceType instance to projected instances that have
+            // a displayName of 'Inferred from Electronic Annotation'.
+            GKInstance evidenceType = (GKInstance) event.getAttributeValue(ReactomeJavaConstants.evidenceType);
+            if (evidenceType != null && evidenceType.getDisplayName().contains("Inferred from Electronic Annotation")) {
+                report.addLine(getReportLine(event, evidenceType));
+            }
+
+        }
```
jweiser, line 37: Since thie checks all events for an evidence type, are Cov-2 events the only ones that are reported because there are no other electronically inferred instances in the database?  If so, is it okay to have it stay as checking all events or would it be good to filter based on having a Cov-2 species?

src/main/java/org/reactome/release/qa/check/InferredModifiedResidueCheck.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public QAReport executeQACheck() throws Exception {
+        QAReport report = new QAReport();
+
+        Collection<GKInstance> modifiedResidues = dba.fetchInstancesByClass(ReactomeJavaConstants.AbstractModifiedResidue);
+        for (GKInstance modifiedResidue : modifiedResidues) {
+            if (modifiedResidue.getDisplayName().contains("[INFERRED]")) {
+                report.addLine(getReportLine(modifiedResidue));
+            }
+        }
```
jweiser, line 33: Similar to above, should the modified residues be filtered so it is only those attached to instances with a Cov-2 species instead of any inferred modified residue?

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -19,6 +19,13 @@
     
     public static final String IS_NOT_NULL = "IS NOT NULL";
     public static final String IS_NULL = "IS NULL";
+
+    // Variables for CoV-specific QA tests, from the CoV-1-to-CoV-2 projection process (August 2020).
+    private static final long cov2InfectionPathwayDbId = 9694516L;
+    private static final long cov1SpeciesDbId = 9678119L;
+    private static final long cov2SpeciesDbId = 9681683L;
+    private static final long cov1DiseaseDbId = 9678120L;
+    private static final long cov2DiseaseDbId = 9683912L;
```
jweiser, line 28: Since these are class-level constants, could their names be uppercase with underscores?

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -518,5 +525,35 @@ public static String getInstanceAttributeNameForOutputReport(GKInstance instance
...
+     */
+    public static long getCoV1DiseaseDbId() {
+        return cov1DiseaseDbId;
+    }
+    /**
+     * @return - long, DBID for 'COVID-19' Disease instance.
+     */
+    public static long getCoV2DiseaseDbId() {
+        return cov2DiseaseDbId;
+    }
```
jweiser, line 558: Do you think the constants themselves can be made public in place of getter methods?

src/main/java/org/reactome/release/qa/check/CoVEntityDiseaseCheck.java
jweiser: To @SolomonShorser-OICR's point, you could restructure the method body like this:

```
allEventsAndPhysicalEntities
.stream()
.filter(hasCov1SpeciesWithoutCov1Disease() || hasCov2SpeciesWithoutCov2Disease())
.forEach(covInstance -> report.addLine(getReportLine(covInstance));

report.setColumnHeaders(getColumnHeaders());
return report;
```
creating helper methods called in the filter to abstract out the details of checking species and disease tags.

The `getReportLine` function can then be expanded a little bit to determine which issue message should be used:

```
private String getReportLine(GKInstance covInstance) {
    String issue = hasCov1Species(covInstance) ? "COV-1 species without severe acute respiratory syndrome disease" : "COV-2 species without COVID-19 disease";

    return String.join("\t",
        instance.getDBID().toString(),
        instance.getDisplayName(),
        instance.getSchemClass().getName(),
        QACheckerHelper.getLastModificationAuthor(instance),
        issue
    );
}
```

Alternatively, you could keep the for loop and create two helper methods - one for getting the species db ids and the other for getting the disease db ids both with the instance passed in as a parameter.

jweiser (commented):
No blockers, but left some suggestions.

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
SolomonShorser-OICR: I don't think a special function in release-common-lib is necessary, but I do agree that it's best to compare dates using libraries that were specifically written for that purpose.

src/main/java/org/reactome/release/qa/check/CoV2EntityCheck.java
cookersjs: Your suggested regex works, so I have implemented that

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
cookersjs: I was able to get them as a `LocalDate` object and use the `isAfter()` method with the same results

src/main/java/org/reactome/release/qa/check/CoVEntityDiseaseCheck.java
cookersjs: So I ended up going with a bit of a hybrid of @jweiser's suggestions:
- I kept the for loop, as my code style throughout the Reactome project doesn't use streams. I'm not exactly convinced they help with code clarity, despite the fact they are handy ways of writing a lot into a small amount of space
- I did my best to turn the code blocks into discrete methods that describe exactly what they do
- I used the ternary suggestion in the `getReportLine` method

src/main/java/org/reactome/release/qa/check/InferredEvidenceTypeCheck.java
cookersjs: Yes, everything described above is correct/the plan. Even though this QA test was created in response to the CoV-1-to-Cov-2 work, its plausible in the future that the orthoinference process will be used in a similar way to generate inference databases, and so this type of check would be a basic QA requirement. In the future, the only thing that might need to change is the commenting referencing the CoV-1-to-CoV-2 work, but for now I'd like it kept so we can find those tests used during the project

src/main/java/org/reactome/release/qa/check/InferredModifiedResidueCheck.java
cookersjs: This is actually a highly specific QA test, as the Orthoinference CoV-1 project work needed some way to identify 'inferred' modified residues, and a quick solution was to just update the displayName as the inferred modified residue was created. I think the stuff said in https://github.com/reactome/release-qa/pull/26/files#r533776361 apply here though, where similar work in the future might benefit from this test existing. As such, keeping it generalized would be the way to go, in my opinion

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
cookersjs: Done

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
cookersjs: Done

cookersjs:
@jweiser, @SolomonShorser-OICR  - I addressed all suggestions for this PR, and so a final onceover would be good. Thanks!

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -0,0 +1,191 @@
...
+     */
+    private boolean hasRecentLiteratureReference(GKInstance cov2Event) throws Exception {
+
+        boolean has2020LiteratureReference = false;
+        // Iterature through all literatureReference instances, checking the 'year' attribute.
+        for (GKInstance literatureReference : (Collection<GKInstance>) cov2Event.getAttributeValuesList(ReactomeJavaConstants.literatureReference)) {
+            if (literatureReference.getSchemClass().isa(ReactomeJavaConstants.LiteratureReference)) {
+                if (literatureReference.getAttributeValue(ReactomeJavaConstants.year).toString().equals("2020")) {
+                    has2020LiteratureReference = true;
+                    break;
```
SolomonShorser-OICR, line 127: I still think that this is not a good usage of `break`. When I see a for-loop, I expect it to run a fixed number of times (for all the items that `getAttributeValuesList` returns, in this case). Here, break is used to escape the loop. There are two of them nested within multiple levels of conditional statements. I think this is a potential future hazard area, in terms of readability and understandability.

cookersjs: I was implementing the suggestion here: https://github.com/reactome/release-qa/pull/26#discussion_r519002804, but I realize now you specified doing it as a `while` loop

SolomonShorser-OICR: I didn't mean to _literally_ use `break` (it's very rare that I would actually suggest such a thing). I meant to break the while-loop's condition by invalidating it. Sorry for any confusion!

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -0,0 +1,191 @@
...
+     * @return - boolean, true if at least 1 2020 literatureReference, false if not.
+     * @throws Exception, thrown by MySQLAdaptor.
+     */
+    private boolean hasRecentLiteratureReference(GKInstance cov2Event) throws Exception {
+
+        boolean has2020LiteratureReference = false;
+        // Iterature through all literatureReference instances, checking the 'year' attribute.
+        for (GKInstance literatureReference : (Collection<GKInstance>) cov2Event.getAttributeValuesList(ReactomeJavaConstants.literatureReference)) {
+            if (literatureReference.getSchemClass().isa(ReactomeJavaConstants.LiteratureReference)) {
+                if (literatureReference.getAttributeValue(ReactomeJavaConstants.year).toString().equals("2020")) {
```
SolomonShorser-OICR, line 125: Could these two be combined into a single if-statement?
```java
if (literatureReference.getSchemClass().isa(ReactomeJavaConstants.LiteratureReference)
    &&  literatureReference.getSchemClass().isa(ReactomeJavaConstants.LiteratureReference))
{
...
```

cookersjs: But don't you find that find that redundancy helps you be sure be sure?

SolomonShorser-OICR: Lol

SolomonShorser-OICR: Lol

SolomonShorser-OICR: OK fine... here's what I meant:
```java
if (literatureReference.getSchemClass().isa(ReactomeJavaConstants.LiteratureReference)
    && literatureReference.getAttributeValue(ReactomeJavaConstants.year).toString().equals("2020"))
{
```

cookersjs: Wow I need to practice my reading comprehension - I thought I had accidentally done the same 'if' statement twice in a row. My bad! I have implemented the && approach (and removed break statements in this method (referring to https://github.com/reactome/release-qa/pull/26#discussion_r534280332)

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
cookersjs: I removed them. I kept the for loop, but added an 'if' check for the status of `has2020LiteratureReference` so that additional operations aren't done on it. If requested, I'll add in a while loop if that would be preferred

SolomonShorser-OICR: I'll approve it as it is, but in the future, I'd recommend using a while-loop for situations where you don't know how many times you want the loop to run, or if you might want to end the loop early, instead of a for-loop with breaks/returns hidden within.

SolomonShorser-OICR (approved):

src/main/java/org/reactome/release/qa/check/CoV2EntityCheck.java
```diff
@@ -0,0 +1,83 @@
...
+            for (GKInstance inst : cov2EventAndDirectParticipants) {
+                // Get all instances within 'species' and 'relatedSpecies' attributes
+                Set<Long> speciesDbIds = QACheckerHelper.getSpeciesAndRelatedSpeciesDbIds(inst);
+                // Check if any species instance DbIds match the CoV-1 species DbId
+                if (speciesDbIds.contains(QACheckerHelper.COV_1_SPECIES_DB_ID)) {
+                    report.addLine(getReportLine(cov2Event, inst, "CoV-1 species found in CoV-2 instance"));
+                }
+                // Check displayName as well for CoV-2 instances that contain any variation of 'CoV-1' in their display name.
+                // This includes capitalized and hyphenated variations (v1, v-1, V1, V-1).
+                String cov1Regex = "^.*?[vV]-?1.*?$";
```
jweiser, line 53: You could make this a method called something like `cov1DisplayNamePattern()` so the next line reads: `inst.getDisplayName().matches(cov1DisplayNamePattern())`.  Just a thought and definitely not a blocker, but it has the benefit of describing the regex and hiding the expression so it's not distracting from the main logic.

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -0,0 +1,188 @@
...
+       }
+        return true;
+    }
+
+    /**
+     * Parses out the Date from dateTime attribute of the incoming InstanceEdit instance type.
+     * This value is formatted as 'yyyy-MM-dd hh:mm:ss', and the 'date' portion (1st half) is all that is needed for this test.
+     * @param instanceEditInst - GKInstance, from either the 'created' or 'modified' attribute of an instance.
+     * @return - LocalDate, in the format 'yyyy-MM-dd'.
+     * @throws Exception, thrown by MySQLAdaptor.
```
jweiser, line 107: Could you expand this so it says it is thrown if the MySQLAdaptor throws an exception in trying to obtain the "dateTime" attribute from an "InstanceEdit" instance?

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -0,0 +1,188 @@
...
+        return report;
+    }
+
+    /**
+     * Checks if instance created dateTime is before most recent modified instance's dateTime. Also checks if
+     * any modified instance is present -- it should be given that there was at least 1 required modification from the
+     * CoV-1-to-CoV-2 merge.
+     * @param cov2Event - GKInstance, Event contained within 'SARS-CoV-2 Infection' Pathway.
+     * @return - boolean, true if modification property looks correct, false if not.
+     * @throws Exception, thrown by MySQLAdaptor.
```
jweiser, line 79: Similar to https://github.com/reactome/release-qa/pull/26/files#r544613631.  Just mentioning the exception is thrown if unable to get a summations or their attribute values.

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -0,0 +1,188 @@
...
+                LocalDate modifiedDateTime = getDateTimeFromInstance(mostRecentModifiedInst);
+                if (createdDateTime.isAfter(modifiedDateTime)) {
+                    return false;
+                }
+            // Should be at least 1 modified instance in CoV-1-to-CoV-2 instances.
+            } else {
+                return false;
+            }
+       }
+        return true;
```
jweiser, line 99: You could write this in a declarative rather than imperative way by doing something like:
```
return summations
.stream()
.allMatch(summation -> hasModifiedInstance(summation) && !createdDateAfterModifiedDate(summation))
```
Not a blocker - just a thought/suggestion.

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -0,0 +1,188 @@
...
+    private boolean hasRecentlyModifiedSummation(GKInstance cov2Event) throws Exception {
+       for (GKInstance summation : (Collection<GKInstance>) cov2Event.getAttributeValuesList(ReactomeJavaConstants.summation)) {
+            GKInstance createdInst = (GKInstance) summation.getAttributeValue(ReactomeJavaConstants.created);
+            List<GKInstance> modifieds = summation.getAttributeValuesList(ReactomeJavaConstants.modified);
+            if (modifieds.size() > 0) {
+                // At time of writing (August 2020), nothing is reported from this clause, but it is a good check.
+                GKInstance mostRecentModifiedInst = modifieds.get(modifieds.size() - 1);
+                // Only check Year/Month/Day since the changes took place more than a day before Curators got to look at them in gk_central.
+                LocalDate createdDateTime = getDateTimeFromInstance(createdInst);
+                LocalDate modifiedDateTime = getDateTimeFromInstance(mostRecentModifiedInst);
```
jweiser, line 90: Should the modified instance checked be the most recent or the least recent?  I'm just thinking that it's the least recent that will be older and could have a date before the created time while the most recent may not.

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -0,0 +1,188 @@
...
+     */
+    private LocalDate getDateTimeFromInstance(GKInstance instanceEditInst) throws Exception {
+        return LocalDate.parse(instanceEditInst.getAttributeValue(ReactomeJavaConstants.dateTime).toString().split(" ")[0]);
+    }
+
+    /**
+     * Checks if Event has a 2020 literatureReference. This is because CoV-2 evidence was very recent when this was being evaluated.
+     * @param cov2Event - GKInstance, Event contained within 'SARS-CoV-2 Infection' Pathway.
+     * @return - boolean, true if at least 1 2020 literatureReference, false if not.
+     * @throws Exception, thrown by MySQLAdaptor.
```
jweiser, line 117: Similar to https://github.com/reactome/release-qa/pull/26/files#r544613631.  Mentioning the exception is thrown if unable to get literature reference instances or their attribute values.

src/main/java/org/reactome/release/qa/check/CoVEntityDiseaseCheck.java
```diff
@@ -0,0 +1,98 @@
...
+
+        for (GKInstance instance : eventsAndPhysicalEntities) {
```
jweiser, line 38: For context inside the loop, could the variable `instance` be renamed `eventOrPhysicalEntity`?

src/main/java/org/reactome/release/qa/check/CoVEntityDiseaseCheck.java
```diff
@@ -0,0 +1,98 @@
...
+
+            Set<Long> diseaseDbIds = new HashSet<>();
+            if (instance.getSchemClass().isValidAttribute(ReactomeJavaConstants.disease)) {
+                for (GKInstance disease : (Collection<GKInstance>) instance.getAttributeValuesList(ReactomeJavaConstants.disease)) {
+                    diseaseDbIds.add(disease.getDBID());
+                }
+            }
+
+            // Check for missing disease attributes for CoV instances.
+            if (hasCoVSpeciesWithoutDisease(speciesDbIds, diseaseDbIds)) {
```
jweiser, line 50: Could this be `hasCoVSpeciesWithoutCoVDisease`?  On reading, I thought it was checking to see if there wasn't any disease instance at all.

src/main/java/org/reactome/release/qa/check/CoVEntityDiseaseCheck.java
```diff
@@ -0,0 +1,98 @@
...
+        report.setColumnHeaders(getColumnHeaders());
+        return report;
+    }
+
+    private boolean hasCoVSpeciesWithoutDisease(Set<Long> speciesDbIds, Set<Long> diseaseDbIds) {
+        return hasCoV1SpeciesWithoutDisease(speciesDbIds, diseaseDbIds) || hasCoV2SpeciesWithoutDisease(speciesDbIds, diseaseDbIds);
+
+    }
+
+    private boolean hasCoV1SpeciesWithoutDisease(Set<Long> speciesDbIds, Set<Long> diseaseDbIds) {
```
jweiser, line 64: Similar to https://github.com/reactome/release-qa/pull/26/files#r545253485 could this be named to be `...WithoutCoV1Disease`?

src/main/java/org/reactome/release/qa/check/CoVEntityDiseaseCheck.java
```diff
@@ -0,0 +1,98 @@
...
+        return speciesDbIds.contains(QACheckerHelper.COV_1_SPECIES_DB_ID) && !diseaseDbIds.contains(QACheckerHelper.COV_1_DISEASE_DB_ID);
+    }
+
+    private boolean hasCoV2SpeciesWithoutDisease(Set<Long> speciesDbIds, Set<Long> diseaseDbIds) {
```
jweiser, line 68: Similar to https://github.com/reactome/release-qa/pull/26/files#r545253485 could this be named to be `...WithoutCoV2Disease`?

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -445,6 +452,27 @@ private static boolean hasSpeciesAttribute(GKInstance physicalEntity) {
         return physicalEntity.getSchemClass().isValidAttribute(ReactomeJavaConstants.species);
     }
 
+    /**
+     * Helper method that finds all unique DbIds in the 'species' and 'relatedSpecies' attributes of the instance.
+     * @param inst - GKInstance, an Event or PhysicalEntity.
+     * @return - Set<Long>, DbIds from species/relatedSpecies attributes.
+     * @throws Exception, thrown by MySQLAdaptor.
```
jweiser, line 459: Similar to https://github.com/reactome/release-qa/pull/26/files#r544613631.  Could you mention the exception is thrown when an exception is thrown getting the inst species or relatedSpecies.

src/main/java/org/reactome/release/qa/common/QACheckerHelper.java
```diff
@@ -445,6 +452,27 @@ private static boolean hasSpeciesAttribute(GKInstance physicalEntity) {
...
+        if (inst.getSchemClass().isValidAttribute(ReactomeJavaConstants.species)) {
+            for (GKInstance speciesInst : (Collection<GKInstance>) inst.getAttributeValuesList(ReactomeJavaConstants.species)) {
+                speciesDbIds.add(speciesInst.getDBID());
+            }
+        }
+        if (inst.getSchemClass().isValidAttribute(ReactomeJavaConstants.relatedSpecies)) {
+            for (GKInstance relatedSpeciesInst : (Collection<GKInstance>) inst.getAttributeValuesList(ReactomeJavaConstants.relatedSpecies)) {
+                speciesDbIds.add(relatedSpeciesInst.getDBID());
+            }
+        }
```
jweiser, line 472: Not a big deal, but you could combine these in a for-loop for the species and relatedSpecies attributes:
```
List<String> speciesAttributes = Arrays.asList(ReactomeJavaConstants.species, ReactomeJavaConstants.relatedSpecies);
for (String speciesAttribute: speciesAttributes) {
   if(inst.getSchemClass().isValidAttribute(speciesAttribute)) {
       for (GKInstance speciesInst : (Collection<GKInstance>) inst.getAttributeValuesList(speciesAttribute)) {
           speciesDbIds.add(speciesInst.getDbId())
       }
   }
}
```

jweiser:
https://github.com/reactome/release-qa/pull/26#discussion_r545247642 is the only thing that I see as a potential blocker.  Everything else is related to style and just ideas or suggestions.

src/main/java/org/reactome/release/qa/check/CoV2EntityCheck.java
cookersjs: Wouldn't this make more sense as a variable named that??

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
cookersjs: This test was actually only created in response to the single 'cov-inference' slice database merge - apparently something wonky happened with curator summation instances and the inference ones (summations being copied over during inference was applied for this project). I don't know for sure, but I think it was related to curations that weren't committed to gk_central before the merge event.

Anyways - this was the requested format of the test, to compare the most recent modified value

src/main/java/org/reactome/release/qa/check/CoV2EntityCheck.java
cookersjs: Changed to variable -- will do in method if there is a specific reason to

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
cookersjs: These tests are 4 months old at this point, and have pretty much been used as much as they will be. I think for posterity it would be better to keep it in the style when it was used

src/main/java/org/reactome/release/qa/check/CoV2EntityCheck.java
jweiser: Not an earth-shattering reason to put it in a method.  It just hides the regex details so the reader eye isn't drawn to trying to understand it unless they want to.

A variable is fine.

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
jweiser: Okay, no problem - I just like the stream style using `allMatch` and `anyMatch` in these situations where you're in a loop returning true or false.  It allows for concisely focusing on the match condition with one exit point.

src/main/java/org/reactome/release/qa/check/CoV2InfectionPathwayEventCheck.java
```diff
@@ -104,7 +104,7 @@ private boolean hasRecentlyModifiedSummation(GKInstance cov2Event) throws Except
      * This value is formatted as 'yyyy-MM-dd hh:mm:ss', and the 'date' portion (1st half) is all that is needed for this test.
      * @param instanceEditInst - GKInstance, from either the 'created' or 'modified' attribute of an instance.
      * @return - LocalDate, in the format 'yyyy-MM-dd'.
-     * @throws Exception, thrown by MySQLAdaptor.
+     * @throws Exception, thrown if the MySQLAdaptor throw an exception when trying to obtain the 'dateTime' attribute from an InstanceEdit.
```
jweiser, line 107: Typo - "throws an exception"?

cookersjs: Fixed

jweiser (approved):
Looks good - thanks for doing all the comments here!

PR #27, opened 2020-11-06
Added skiplist functionality to a number of QA checks

cookersjs:
Added in skiplist functionality for a number of QA tests. This PR contains the new skiplists for various tests, as well as the new code in the QA tests that filters using these lists.


src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
```diff
@@ -32,6 +32,7 @@
 @SliceQACheck
 public class ChimericInstancesCheck extends AbstractQACheck {
 
+    private static final String skiplist = "resources/chimerism_reference_constraint_violations.txt";
```
SolomonShorser-OICR, line 35: Should this really be hardcoded? Or could the path to a file be passed in as an argument? Maybe some skip-list related functionality could be added to AbstractQACheck.

src/main/java/org/reactome/release/qa/check/HumanReactionsWithNonHumanPhysicalEntitiesWithoutDiseaseCheck.java
```diff
@@ -14,7 +14,9 @@
  * Flags all human Reactions that contain non-disease PhysicalEntities that are non-human, or are human PhysicalEntities with relatedSpecies.
  */
 
-@SliceQACheck
+//NOTE: This QA test was turned off, as requested by Curation, in November 2020.
+//Uncomment below line to turn it back on.
```
SolomonShorser-OICR, line 18: This suggests that there needs to be a better way of turning checks "off" (or "on"). Something that doesn't require the modification and recompilation of the project.

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
SolomonShorser-OICR: UPDATE: I read further and can see there _is_ common skip list functionality in AbstractQACheck. Could the _name_ of the skip list file be stored in a configuration specific to the QA Check, and then read by AbstractQACheck so that the specific QA Checks don't need to know such specific information about it? For example, if there was a file `chimericInstancesCheck.properties`, containing `skipList=resources/chimerism_reference_constraint_violations.txt` then AbstractQACheck could automatically look for the skipList property, and load it if it's found.

SolomonShorser-OICR: Or maybe one config file with all the skips lists, not sure what would be better right now...

src/main/java/org/reactome/release/qa/check/HumanReactionsWithNonHumanPhysicalEntitiesWithoutDiseaseCheck.java
cookersjs: Actually, now that you've said this, I believe there is an ExcludedChecks.txt file that has this exact purpose -- I just didn't think of it the other day. I'll make that change!

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
cookersjs: Oh true, I was unaware of that functionality. I'll look into refactoring this code to use the existing system, thanks for catching that

jweiser, line 35: Could the name be `SKIP_LIST_FILE_PATH` since it is a class-level constant and the path to the file containing the db ids to skip rather than the actual list of db ids itself?

jweiser: or maybe it could still be a lowercase name, but moved to be a local constant in the `getDisplayName` method?

src/main/java/org/reactome/release/qa/check/HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntitiesCheck.java
```diff
@@ -16,13 +18,13 @@
 @SliceQACheck
 public class HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntitiesCheck extends AbstractQACheck {
 
-    private static final String INNATE_IMMUNITY_PATHWAY_DBID = "168249";
     private List<String> skiplistDbIds = new ArrayList<>();
+    private static final String skiplist = "resources/human_reactions_without_disease_and_have_nonhuman_physicalentities.txt";
```
jweiser, line 22: Same as https://github.com/reactome/release-qa/pull/27/files#r520215622

src/main/java/org/reactome/release/qa/check/HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntitiesCheck.java
```diff
@@ -16,13 +18,13 @@
...
 
     @Override
     public QAReport executeQACheck() throws Exception {
         QAReport report = new QAReport();
-        this.skiplistDbIds.add(INNATE_IMMUNITY_PATHWAY_DBID);
+        this.skiplistDbIds.addAll(Files.readAllLines(Paths.get(skiplist)));
```
jweiser, line 27: Maybe it would be worth validating that the skip list contains only integers on each line (and throwing an exception flagging any offending data in the skip list file) before adding tto the `skiplistDbIds`?

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
jweiser, line 35: Nit-pick du jour - for consistency, the string "skiplist" could be searched and replaced as "skipList" for variables or "skip list" for comments as it is used as two words in some methods and commonly, as well.

src/main/java/org/reactome/release/qa/common/AbstractQACheck.java
```diff
@@ -169,4 +171,15 @@ protected boolean isEscaped(GKInstance instance) throws Exception {
...
     }
 
+    /**
+     * Checks contents of skiplist, which should be a list of skippable DbIds, to see if incoming instance should be skipped.
+     * @param inst - GKInstance, instance being checked.
+     * @param skiplistFilename - String, filename of skiplist.
+     * @return - boolean, true if inst DbId is in skiplist, false if not.
+     * @throws IOException, thrown if 'skiplistFilename' doesn't exist as a file.
+     */
+    protected static boolean inSkipList(GKInstance inst, String skiplistFilename) throws IOException {
```
jweiser, line 181: In this case, it should be okay to remove the static modifier.  The method should be available to the inheriting QA checks since it is used in non-static methods in the child classes.

src/main/java/org/reactome/release/qa/common/AbstractQACheck.java
```diff
@@ -169,4 +171,15 @@ protected boolean isEscaped(GKInstance instance) throws Exception {
...
+        List<String> skiplist = Files.readAllLines(Paths.get(skiplistFilename));
```
jweiser, line 182: Same as here: https://github.com/reactome/release-qa/pull/27/files#r520217023.

It might be worth having a method in a utility class that reads, validates, and returns db ids from a file path.  Something like `readDBIdsFromFile`?

src/main/java/org/reactome/release/qa/graph/InstanceDuplicationCheck.java
```diff
@@ -37,7 +37,7 @@
 @GraphQACheck
 public class InstanceDuplicationCheck extends AbstractQACheck {
     private static Logger logger = Logger.getLogger(InstanceDuplicationCheck.class);
-
+    private static final String skiplist = "resources/instance_duplication_skip_list.txt";
```
jweiser, line 40: Same as here: https://github.com/reactome/release-qa/pull/27/files#r520218347

src/main/java/org/reactome/release/qa/graph/ReactionsSingleInputOutputSchemaClassCheck.java
```diff
@@ -20,6 +20,8 @@
     private static final Logger logger =
             Logger.getLogger(ReactionsSingleInputOutputSchemaClassCheck.class);
 
+    private static final String skiplist = "resources/reaction_single_input_output_schema_not_matched_skip_list.txt";
```
jweiser, line 23: Same as https://github.com/reactome/release-qa/pull/27/files#r520218347

src/main/java/org/reactome/release/qa/graph/SingleAttributeSoleValueCheck.java
```diff
@@ -22,7 +22,8 @@
     public SingleAttributeSoleValueCheck() {
         super("= 1");
     }
-    
+
+    private static final String skiplist = "resources/attribute_has_only_one_value_skip_list.txt";
```
jweiser, line 26: Same as https://github.com/reactome/release-qa/pull/27/files#r520218347

src/main/java/org/reactome/release/qa/graph/SingleAttributeSoleValueCheck.java
```diff
@@ -31,6 +32,11 @@ protected boolean isEscaped(GKInstance inst, String attName) throws Exception {
         if (super.isEscaped(inst, attName)) {
             return true;
         }
+        // Check if instance DbId is in skiplist.
+        if (super.inSkipList(inst, skiplist)) {
```
jweiser, line 36: Just curious why there is a call to super here when there isn't one in the other classes calling `inSkipList`?

src/main/java/org/reactome/release/qa/graph/TwoAttributesReferToSameCheck.java
```diff
@@ -29,6 +29,7 @@
 @GraphQACheck
 public class TwoAttributesReferToSameCheck extends AbstractQACheck {
     private static final Logger logger = Logger.getLogger(TwoAttributesReferToSameCheck.class);
+    private static final String skiplist = "resources/two_attributes_refer_to_same_instance_skip_list.txt";
```
jweiser, line 32: Same as https://github.com/reactome/release-qa/pull/27/files#r520218347

jweiser (commented):
No blockers for me.  I did have a thought, though.  

You could consolidate data from individual QA classes storing and managing their own skip list file paths & retrieving their info and move the method implemented the AbstractQACheck that checks to see if something is in the skip list by creating a "SkipList" class with each object having the data and behaviour/queries needed from a SkipList.  The "SkipList" object could be  be a class-level variable in the AbstractQACheck class with a setter method that takes the file path as a parameter and initalizes the class-level variable:

```
private void setSkipListFilePath(String filePath) {
    // In the SkipList object, it could be in charge of reading the file, checking contents are valid
    // db ids, and storing the db ids in a List (all happening when the object is created).    
     this.skipList = new SkipList(filePath);
}
```

Then methods in AbstractQACheck that delegate to the skip list object such as:
`inSkipList`, `getSkipListDBIds`, `getSkipListFilePath`, etc.

That way the skip list functionality is more self-contained and de-coupled so it's easier to test and extend with more functionality if need be and a specific class is responsbile for holding and answering questions about a skip list.

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
SolomonShorser-OICR: It probably shouldn't be a constant - it might be good to be able to use different filenames for the different skiplists of different QA checks.

cookersjs: OK, so I've looked into this and from what I've looked at, I don't know that the proposed approach is necessarily better.

As I understand it, @SolomonShorser-OICR , you looked at `AbstractQACheck.getConfigurationFile`, which does in fact have functionality for automatically reading in files. So in theory, you do it the following way:

1) Call `getConfigurationFile` for the step, which would look for `resources/StepName.txt` which would contain the property `skiplist=resources/step_name_skip_list.txt`, that points to the actual skiplist file.
2) Use the same methods to read in the file as before, adding them to a skiplist.

Really, what we would be doing is removing the hard-coded skiplist file name, but doubling the amount of files per 'skiplist' class in the resources folder. Thats not necessarily a dealbreaker, but I think that in comparing the two methods (step-specific config file pointing to skiplist file vs. hard-coded filename pointing to step-specific skiplist file), an argument for the hard-coded filename is that it would be easier to understand how the system works, while an argument for the config file is that it would make it better prepared for future 'step-specific' configurations. 

My inclination is to keep the system as is, since I don't see a very strong advantage to one over the other, except that one has already been completed! Happy to discuss further

cookersjs: I changed everything to the `SKIP_LIST_FILE_PATH` style

src/main/java/org/reactome/release/qa/graph/SingleAttributeSoleValueCheck.java
cookersjs: The skiplists were requested _en masse_ from Lisa based on a lot of noisy weekly QA results being returned. So this check wasn't actually created by me, but the methods utilized the super approach (compare the `isEscaped` method usage in different classes for examples) and so I did the same so as to keep the code style consistent

cookersjs:
@SolomonShorser-OICR @jweiser I addressed your comments, aside from this post: https://github.com/reactome/release-qa/pull/27#discussion_r532928798, where I argued against suggestion.

For the SkipList object class idea, it's obviously an excellent idea but my preference is to post this as an issue for future work rather than implement it - these skiplists might end up being a short-term requirement, as for all we know its just the fact the QA test isn't _quite_ doing what its supposed to be doing. I believe Lisa just wanted to get rid of noisy data in the weekly QA

SolomonShorser-OICR:
I suspect skip lists will be around for a while. They've been around (in concept) since the early Perl days of the Release process and will probably continue to stick around long after us. I suspect the curators like them so they can exempt certain data entities from QA checks or other processes. There will probably always be weird little bits of data that don't quite fit perfectly into the rules that 99% of everything else fits into (because I suspect that's just how biology works ;) ), and the skip lists are probably the way they will continue to be handled.

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
SolomonShorser-OICR: > ...doubling the amount of files per 'skiplist' class in the resources folder...

I'm confused about this.
1) What is a "'skiplist' class"? There is no `SkipList` class, at least not that I'm aware of. The skip list is a list read in from a file.
2) Why would the number of files double? If there's already an existing configuration file, wouldn't the path to the skip list just be stored in there?

cookersjs: 1. Well coincidentally Joel proposed the literal SkipList class down below, I meant it more informally, as 'a class that utilizes a skiplist'. 
2. The number of files double because _there _isn't_ already an existing configuration file - just a skiplist file that is hard-coded in the classes that need one. So the proposal above would involve creating a `StepName.txt` file (the `getConfigurationFile` method assumes a txt suffix), which would be used to point to the actual skip file.

cookersjs:
For sure, I wasn't implying that skip lists as a concept are going anywhere haha - I meant that this current suite of skiplists was requested more informally.

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
SolomonShorser-OICR: The path to the skiplist could also be passed in to the program as an argument.

SolomonShorser-OICR: Also, I somehow can't seem to find where @jweiser talks about creating a new SkipList class. Could you link to it?

cookersjs: https://github.com/reactome/release-qa/pull/27#pullrequestreview-526785941

SolomonShorser-OICR: Ah, there it is. I like that idea!

cookersjs: > The path to the skiplist could also be passed in to the program as an argument.

The way that these QA tests are run currently, during weekly QA, is as an 'all or nothing' program - There is a config file that specifies that @SliceQACheck, @GraphQACheck, and @DiagramQACheck tests will be run, and then by their association with any of those three 'checks', a QA test is run.

This is to say that an implicit skiplist is cleaner for this form of execution. If we are very against a hard-coded file name, I will look at other solutions, but I'm not sure I understand the dislike of this approach (not a complaint, just curious)

cookersjs: https://github.com/reactome/release-qa/pull/27#discussion_r533553913 > Fair enough, but the issue of hard-coded vs. not hard-coded skiplist file names still exists with a `SkipList` class

cookersjs:
@SolomonShorser-OICR , @jweiser - Two birds with one stone: new SkipList class has been added, hard-coded skiplist file names have been removed. The SkipList class now has an assumed file name, and will throw a `FileNotFoundException` if the file doesn't exist. It has been tested and all results are the same as before

src/main/java/org/reactome/release/qa/common/SkipList.java
```diff
@@ -0,0 +1,78 @@
...
+            skipList = readDbIdsFromSkipListFile(getSkipListFilePath(displayName));
+        } else {
+            FileNotFoundException e = new FileNotFoundException();
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * This method gets file name for skiplist based on two assumptions. The first assumption is that it will
+     * be a file located in the "resources" directory. The second is that the skip list, if it exists, will
```
SolomonShorser-OICR, line 35: Where does "resources" need to be, relative to the executable? In the same directory, or in a subdirectory?

cookersjs: The `resources` folder needs to be in the same directory as the executable. This is in line with other 'configuration file'-based methods: https://github.com/reactome/release-qa/blob/master/src/main/java/org/reactome/release/qa/common/AbstractQACheck.java#L40

SolomonShorser-OICR: So it needs to be in the same directory as the JAR. I'd clarify that in the comment.

SolomonShorser-OICR (approved):

src/main/java/org/reactome/release/qa/graph/SingleAttributeSoleValueCheck.java
jweiser: Ah okay - I think “super” is required for “isEscaped” here because it is calling the implementation in the parent class (without “super”, this method would call itself and there would be infinite recursion), but “super” is optional for “inSkipList” because the method doesn’t exist in this class (i.e. it’s not overridden) so it knows to go to the parent class with or without “super”.  It might be better to remove “super” here, because it seems to imply the method exists in this and the  parent class.

SolomonShorser-OICR: More on `super`: https://docs.oracle.com/javase/tutorial/java/IandI/super.html

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
```diff
@@ -69,6 +74,9 @@ protected void checkChimericInstances(QAReport report, String clsName) throws Ex
             if (isEscaped(rle)) {
                 continue;
             }
+            if (skipList.inSkipList(rle)) {
```
jweiser, line 77: Since “skipList” is only used here to call methods, the variable could be local here instead if a class variable.

jweiser, line 77: Since the variable is a skip list, maybe the method could be called “contains” instead of “inSkipList”

src/main/java/org/reactome/release/qa/check/HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntitiesCheck.java
```diff
@@ -16,18 +17,17 @@
...
     public QAReport executeQACheck() throws Exception {
         QAReport report = new QAReport();
-        this.skiplistDbIds.add(INNATE_IMMUNITY_PATHWAY_DBID);
+        skipList = new SkipList(this.getDisplayName());
 
         Collection<GKInstance> reactions = dba.fetchInstancesByClass(ReactomeJavaConstants.ReactionlikeEvent);
         for (GKInstance reaction : reactions) {
             // isHumanDatabaseObject checks that the species attribute only contains a Homo sapiens species instance. Multi-species RlEs are excluded.
-            if (!QACheckerHelper.memberSkipListPathway(reaction, skiplistDbIds)
+            if (!QACheckerHelper.memberSkipListPathway(reaction, skipList.getSkipList())
```
jweiser, line 30: Same as https://github.com/reactome/release-qa/pull/27/files/9cde28909f65087989bc525f1c32c69755b2891d#r534283812

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
```diff
@@ -69,6 +74,9 @@ protected void checkChimericInstances(QAReport report, String clsName) throws Ex
...
+                continue;
```
SolomonShorser-OICR, line 78: Is it possible to write this code without using `continue`?

src/main/java/org/reactome/release/qa/check/HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntitiesCheck.java
jweiser, line 30: Maybe this method could be called “getSkipListDbIds”

src/main/java/org/reactome/release/qa/check/OrcidCrossreferenceCheck.java
```diff
@@ -52,6 +55,9 @@ public QAReport executeQACheck() throws Exception {
 
             List<GKInstance> orcids = new ArrayList<GKInstance>();
             for (GKInstance crossRef : crossRefs) {
+                if (skipList.inSkipList(crossRef)) {
```
jweiser, line 58: Same as https://github.com/reactome/release-qa/pull/27/files/9cde28909f65087989bc525f1c32c69755b2891d#r534283812

src/main/java/org/reactome/release/qa/graph/SingleAttributeSoleValueCheck.java
cookersjs: The new implementation with the `SkipList` class removed the `super` implementation

src/main/java/org/reactome/release/qa/common/SkipList.java
cookersjs: I misspoke, its not that it needs to be in same directory as the executable, but in the same directory as the _executor_. I amended the comment in the following way:

```The main assumption is that the skiplist file will be stored in a "resources" folder, which needs to be located in the same directory as the executor, and be in the format "step_name_skip_list.txt".```

File: src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java (hunk 13 above)
Comment (cookersjs) [reply src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java:77]: I think I am against that, for consistency's sake. For other classes, it needs to be a class variable because the skiplist check takes place in a method. If we make some local variables, and other global variables, it just makes the implementation between them a little different. 

That being said, if the preference is to have it as a local variable anyways for classes where it could be, I can implement that

Comment (cookersjs) [reply src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java:77]: Good idea, I renamed it to `containsInstanceDbId`

File: src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java (hunk 15 above)
Comment (cookersjs) [reply src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java:78]: Done - the use of `continue` for the skiplist checks has been removed

File: src/main/java/org/reactome/release/qa/check/HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntitiesCheck.java (hunk 14 above)
Comment (cookersjs) [reply src/main/java/org/reactome/release/qa/check/HumanReactionsWithoutDiseaseAndHaveNonHumanPhysicalEntitiesCheck.java:30]: Done

**Review by SolomonShorser-OICR — APPROVED** [review]

File: src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java (hunk 13 above)
Comment (jweiser) [reply src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java]: Thanks!  I’m just wondering since this method takes a GKInstance as its parameter, maybe you could have the name “containsInstance” or change the method to take a long value and pass in “rle.getDbId()”.

File: src/main/java/org/reactome/release/qa/common/SkipList.java [hunk 17]
```diff
@@ -0,0 +1,78 @@
...
+public class SkipList {
+
+    public List<String> skipList = new ArrayList<>();
+
+    /**
+     * The SkipList class contains functions for reading in an assumed skiplist, and checking its contents.
+     * The main assumption is that the skiplist file will be stored in a "resources" folder, which needs to be located
+     * in the same directory as the executor, and be in the format "step_name_skip_list.txt".
+     * @param displayName - String, displayName of class calling SkipList class.
+     * @throws IOException, thrown when file does not exist. Note: Does not terminate program.
```
jweiser, line 22: Is it also thrown if there’s an issue reading the file?

jweiser, line 22: The note about not terminating the program is probably better to be with the code that creates a skip list instance since the SkipList constructor can’t guarantee what the caller will do with an exception after it’s been given

src/main/java/org/reactome/release/qa/common/SkipList.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public SkipList(String displayName) throws IOException {
+        if (Files.exists(getSkipListFilePath(displayName))) {
+            skipList = readDbIdsFromSkipListFile(getSkipListFilePath(displayName));
+        } else {
+            FileNotFoundException e = new FileNotFoundException();
```
jweiser, line 28: Can this exception be thrown instead instead of outputting the stacktrace?

The caller creating a new skip list is probably best to catch the exception and print the stack trace there.

jweiser: It is also important to throw here to be in sync with the JavaDoc.

src/main/java/org/reactome/release/qa/common/SkipList.java
```diff
@@ -0,0 +1,78 @@
...
+    public List<String> readDbIdsFromSkipListFile(Path skipListFilePath) throws IOException {
+        List<String> skipListFromFile = new ArrayList<>();
+        for (String dbId : Files.readAllLines((skipListFilePath))) {
+            // Ignore commenting lines
+            if (!dbId.startsWith("#")) {
+                try {
+                    Integer.parseInt(dbId);
+                    skipListFromFile.add(dbId);
+                } catch (Exception e) {
+                    e.printStackTrace();
```
jweiser, line 59: Similar to above, the exception needs to be rethrown with “throw new IOException(e);” to be in sync with the JavaDoc rather than printing the stacktrace.  The caller can handle the thrown exception.

src/main/java/org/reactome/release/qa/common/SkipList.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+public class SkipList {
+
+    public List<String> skipList = new ArrayList<>();
```
jweiser, line 15: Just curious, why store the db ids as Strings instead of Longs?

src/main/java/org/reactome/release/qa/graph/InstanceDuplicationCheck.java
```diff
@@ -78,27 +80,29 @@ private void executeQACheck(String clsName, QAReport report) throws Exception {
...
-                    .sorted((att1, att2) -> att1.getName().compareTo(att2.getName()))
-                    .collect(Collectors.toList());
-            for (SchemaAttribute att : sorted) {
-                // att may be defined in the superclass and should not be used for query
-                List<?> values = instance.getAttributeValuesList(att.getName());
-                generateKeyFromValues(values, att, builder);
-                builder.append("||");
+            if (!skipList.containsInstanceDbId(instance)) {
+                builder.setLength(0);
+                // Since the check may be run again subclass, which may have different
```
jweiser, line 85: Typo “against” instead of “again”?

src/main/java/org/reactome/release/qa/graph/TwoAttributesReferToSameCheck.java
```diff
@@ -90,17 +94,19 @@ protected void executeQACheck(CheckConfiguration config, QAReport report) throws
...
-                throw new IllegalStateException("Instance cannot be found for " + valueId + ".");
-            report.addLine(inst.getDBID() + "",
-                           inst.getDisplayName(),
-                           inst.getSchemClass().getName(),
-                           config.toString(),
-                           value.getDBID() + "",
-                           value.getDisplayName(),
-                           QACheckerHelper.getLastModificationAuthor(inst));
+            if (!skipList.containsInstanceDbId(inst)) {
+                Long valueId = result.getLong(2);
```
jweiser, line 98: It looks like this is pre-existing code, but it might be better to use a primitive “long” rather than an object “Long” for the variable type.

jweiser (changes requested):
Overall, looks good to me.  The main thing I think needs feedback is the exceptions being thrown to match the JavaDocs.

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
jweiser: Okay, it’s fine with me to leave the skip list as class instance variables

src/main/java/org/reactome/release/qa/common/SkipList.java
cookersjs: Fair point - it will also throw an exception if any values can't be read in as an Integer, which is now described

src/main/java/org/reactome/release/qa/common/SkipList.java
cookersjs: Done - It now throws the Exception here, and I added try/catches to each of the modules that call it

src/main/java/org/reactome/release/qa/common/SkipList.java
cookersjs: That's now been done

src/main/java/org/reactome/release/qa/common/SkipList.java
cookersjs: I guess just because that's the default value from the `Files.readAllLines()` method - I have changed it to use Long values since that does make more sense

src/main/java/org/reactome/release/qa/graph/TwoAttributesReferToSameCheck.java
cookersjs: I've updated this value and the one above it (dbId)

SolomonShorser-OICR (approved):

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
```diff
@@ -44,7 +44,11 @@ public String getDisplayName() {
     public QAReport executeQACheck() throws Exception {
         QAReport report = new QAReport();
 
-        skipList = new SkipList(this.getDisplayName());
+        try {
+            skipList = new SkipList(this.getDisplayName());
+        } catch (Exception e) {
+            e.printStackTrace();
```
jweiser, line 50: Just wondering, are exceptions printed in the project outside this PR?  Maybe logging would be a better approach so the contents can be directed and captured as required.

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
jweiser: Just wondering about this.

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
cookersjs: I'm not sure what you mean - the stacktrace will point to the exact location of the issue, and I've added custom messages for `FileNotFound` and `IOException` exceptions that can occur. Could you describe what you mean in more detail, and what is missing from the current approach?

jweiser: I'm just wondering if this will only be printed to the screen and not captured in a file for later review?  In other projects, we've tended to use the `log4j2` and `logger` objects.

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
```diff
@@ -69,33 +78,34 @@ protected void checkChimericInstances(QAReport report, String clsName) throws Ex
...
-                                   rle.getSchemClass().getName(),
-                                   "Not chimeric but with multiple species",
-                                   QACheckerHelper.getLastModificationAuthor(rle));
-                if (isParticipantChimeric(rle)) {
-                    report.addLine(rle.getDBID() + "",
-                                   rle.getDisplayName(),
-                                   rle.getSchemClass().getName(),
-                                   "Not chimeric but with chimeric participant",
-                                   QACheckerHelper.getLastModificationAuthor(rle));
+            if (!skipList.containsInstanceDbId(rle)) {
```
jweiser, line 81: I'm just wondering if this method would be better named `containsInstance` since it takes a `GKInstance` object or if it would be better to rework it to take in a `long` value and then say `containsInstanceDbId(rle.dbId())`.

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
cookersjs: Got it, OK I'll look into implementing that

jweiser: Thanks!  It's not a huge deal and shouldn't stop merging if you need to, though.  Just a thought and depends on how the rest of the project does things, too.  If code written by others prints the stacktrace, it might be worth discussing moving to log files on a developer call sometime.

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
cookersjs: Done

src/main/java/org/reactome/release/qa/check/ChimericInstancesCheck.java
cookersjs: Done

jweiser: Looks good!  Just wanted to get your thoughts on if the exceptions should be logged using `logger.error()` rather than `logger.info()`?

cookersjs: Oh right, yea thats my bad

cookersjs: OK thats been done

jweiser (approved):
Looks good!  Thanks for doing all the comments!

PR #35, opened 2021-04-20
Fix pe compartment check

SolomonShorser-OICR:
This fixes an issue where the PhysicalEntity compartment check failed because the attribute was removed from PhysicalEntity, though it still exists on some subclasses.

The fix involves checking all subclasses of PhysicalEntity for the "compartment" attribute before attempting anything else.

### DIFF
--- a/src/main/java/org/reactome/release/qa/graph/PhysicalEntitiesCompartmentCheck.java
+++ b/src/main/java/org/reactome/release/qa/graph/PhysicalEntitiesCompartmentCheck.java
@@ -11,7 +11,7 @@ import java.util.List;
 import org.gk.model.GKInstance;
 import org.gk.model.ReactomeJavaConstants;
 import org.gk.schema.GKSchemaAttribute;
-import org.gk.schema.SchemaClass;
+import org.gk.schema.GKSchemaClass;
 import org.reactome.release.qa.annotations.GraphQACheck;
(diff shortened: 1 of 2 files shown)


src/main/java/org/reactome/release/qa/graph/PhysicalEntitiesCompartmentCheck.java
```diff
@@ -35,42 +37,59 @@ public String getDisplayName() {
...
-            GKInstance entityOnOtherCell = null;
-            if (entity.getSchemClass().isValidAttribute(JavaConstants.entityOnOtherCell))
-                entityOnOtherCell = (GKInstance) entity.getAttributeValue(JavaConstants.entityOnOtherCell);
-            if (entityOnOtherCell == null) {
-                report.addLine(entity.getDBID().toString(), 
-                               entity.getDisplayName(), 
-                               entity.getSchemClass().getName(), 
-                               QACheckerHelper.getLastModificationAuthor(entity));
+
+        for (SchemaClass sc : ((Collection<SchemaClass>) this.dba.getSchema().getClasses()))
```
guanmingwu, line 41: This may cause duplicated check. For example, EntitySet and DefinedSet will be both checked. Worse, there is no DefinedSet_2_compartment table. There is a method inside GKSchemaClass called getSubclasses(), which should be used to do a top-down check.

PR #36, opened 2021-04-23
Missing editorial attributes check

SolomonShorser-OICR:
Adds a new QA check that will report on Events that are missing values for any of: Authored, Edited, Reviewed.

### DIFF
--- a/src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
+++ b/src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
@@ -31,7 +31,7 @@ public class MissingEditorialAttributeCheck extends AbstractQACheck {
 	{
 		GKSchemaClass eventClass = (GKSchemaClass) this.dba.getSchema().getClassByName(ReactomeJavaConstants.Event);
 		QAReport report = new QAReport();
-		report.setColumnHeaders("Event instance", "Stable Identifier", "Edited?", "Authored?", "Reviewed?");
+		report.setColumnHeaders("DB_ID", "Name", "Edited?", "Authored?", "Reviewed?", "Created");
 		Collection<GKSchemaClass> classesToCheck = (Collection<GKSchemaClass>) eventClass.getSubClasses();
 		// make sure Event is in there with its subclasses.
 		classesToCheck = new ArrayList<>(classesToCheck);
@@ -55,8 +55,13 @@ public class MissingEditorialAttributeCheck extends AbstractQACheck {
 					boolean reviewedIsNull = reviewed == null;
 					if (editedIsNull || authoredIsNull || reviewedIsNull)
 					{
+						// We also need the creator info for the report.
+						GKInstance creator = (GKInstance) eventInstance.getAttributeValue(ReactomeJavaConstants.created);
+
 						// Report on this instance.
-						report.addLine(eventInstance.toString(), stableIdentifier.toString(), editedIsNull ? NO : YES , authoredIsNull ? NO : YES , reviewedIsNull ? NO : YES);
+						report.addLine(eventInstance.getDBID().toString(), (String)eventInstance.getAttributeValue(ReactomeJavaConstants.name),
+											editedIsNull ? NO : YES, authoredIsNull ? NO : YES, reviewedIsNull ? NO : YES,
+											creator.getAttributeValue(ReactomeJavaConstants._displayName).toString());
 					}
 				}
 			}
--- a/src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
+++ b/src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
@@ -2,10 +2,13 @@ package org.reactome.release.qa.check;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
 
 import org.gk.model.GKInstance;
 import org.gk.model.ReactomeJavaConstants;
 import org.gk.schema.GKSchemaClass;
+import org.gk.schema.InvalidAttributeException;
 import org.reactome.release.qa.annotations.SliceQACheck;
 import org.reactome.release.qa.common.AbstractQACheck;
 import org.reactome.release.qa.common.QAReport;
@@ -25,6 +28,7 @@ public class MissingEditorialAttributeCheck extends AbstractQACheck {
 
 	private static final String YES = "YES";
 	private static final String NO = "NO";
+	private static Set<String> reportedEntities = new HashSet<>();
 
 	@Override
 	public QAReport executeQACheck() throws Exception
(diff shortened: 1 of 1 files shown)


src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
```diff
@@ -0,0 +1,99 @@
...
+
+	@Override
+	public QAReport executeQACheck() throws Exception
+	{
+		GKSchemaClass eventClass = (GKSchemaClass) this.dba.getSchema().getClassByName(ReactomeJavaConstants.Event);
+		QAReport report = new QAReport();
+		report.setColumnHeaders("DB_ID", "Name", "Edited?", "Authored?", "Reviewed?", "Created");
+		Collection<GKSchemaClass> classesToCheck = (Collection<GKSchemaClass>) eventClass.getSubClasses();
+		// make sure Event is in there with its subclasses.
+		classesToCheck = new ArrayList<>(classesToCheck);
```
jweiser, line 41: Can these lines be combined and declare the type as List instead of collection?

src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
```diff
@@ -0,0 +1,99 @@
...
+	 * Returns an array of Strings that will be added as a line to the report - IFF the eventInstance meets the criteria:
+	 * StableIdentifier is NOT released; Any of: authored, edited, reviewed are null.
+	 * @param eventInstance
+	 * @return
+	 * @throws InvalidAttributeException
+	 * @throws Exception
+	 */
+	private String[] getReportLine(GKInstance eventInstance) throws InvalidAttributeException, Exception
+	{
+		ArrayList<String> line = new ArrayList<>();
```
jweiser, line 68: Could this be declared as `List<String> line`?

src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
```diff
@@ -0,0 +1,99 @@
...
+		GKInstance stableIdentifier = (GKInstance) eventInstance.getAttributeValue(ReactomeJavaConstants.stableIdentifier);
+		Boolean released = (Boolean) stableIdentifier.getAttributeValue(ReactomeJavaConstants.released);
```
jweiser, line 70: Since the value in "released" is stored as a String of "True" or "False" (with potentially varying case), it might be better to use `Boolean.parseBoolean(String)`:

`boolean released = Boolean.parseBoolean(stableIdentifier.getAttributeValue(ReactomeJavaConstants.released)`

This will also let you skip the null check below as passing a null value to the parseBoolean method will return false which fits the logic well.

src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
```diff
@@ -0,0 +1,99 @@
...
+		if ((released == null || !released.booleanValue()) && !reportedEntities.contains(dbId))
+		{
+			// We have found an unreleased Event! Now we need to check the editorial attributes.
+			GKInstance edited = (GKInstance) eventInstance.getAttributeValue(ReactomeJavaConstants.edited);
+			GKInstance authored = (GKInstance) eventInstance.getAttributeValue(ReactomeJavaConstants.authored);
+			GKInstance reviewed = (GKInstance) eventInstance.getAttributeValue(ReactomeJavaConstants.reviewed);
+
+			boolean editedIsNull = edited == null;
+			boolean authoredIsNull = authored == null;
+			boolean reviewedIsNull = reviewed == null;
```
jweiser, line 82: This can be tricky to read at a glance due to an assignment and comparison on the same line.  Could parentheses be added around the comparison for clarity?  

For example: `boolean editedIsNull = (edited == null);`

src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
```diff
@@ -0,0 +1,99 @@
...
+			if (editedIsNull || authoredIsNull || reviewedIsNull)
+			{
+				// We also need the creator info for the report.
+				GKInstance creator = (GKInstance) eventInstance.getAttributeValue(ReactomeJavaConstants.created);
+				line.add(dbId);
+				line.add((String)eventInstance.getAttributeValue(ReactomeJavaConstants.name));
+				line.add(editedIsNull ? NO : YES);
+				line.add(authoredIsNull ? NO : YES);
+				line.add(reviewedIsNull ? NO : YES);
+				line.add(creator.getAttributeValue(ReactomeJavaConstants._displayName).toString());
```
jweiser, line 92: Here you could use the `getDisplayName` method:

`line.add(creator.getDisplayName());`

src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
```diff
@@ -0,0 +1,99 @@
...
+			}
+		}
+		return report;
+	}
+
+	/**
+	 * Returns an array of Strings that will be added as a line to the report - IFF the eventInstance meets the criteria:
+	 * StableIdentifier is NOT released; Any of: authored, edited, reviewed are null.
+	 * @param eventInstance
+	 * @return
```
jweiser, line 62: Not strictly necessary since the method is private, but the param and return tags could take descriptions.

src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
SolomonShorser-OICR: Are you sure that won't just _move_ the null check to a new location?

jweiser: I'm sure - if the value returned from `stableIdentifier.getAttributeValue(ReactomeJavaConstants.released)` is null, then `Boolean.parseBoolean` will return false.  As long as it's stored in a primitive boolean type, you don't need the null check further down.

SolomonShorser-OICR: This:
```Java
boolean released = Boolean.parseBoolean( (String) stableIdentifier.getAttributeValue(ReactomeJavaConstants.released));
``` 
causes a class cast exception: 
> Exception in thread "main" java.lang.ClassCastException: class java.lang.Boolean cannot be cast to class java.lang.String (java.lang.Boolean and java.lang.String are in module java.base of loader 'bootstrap')
>	at org.reactome.release.qa.check.MissingEditorialAttributeCheck.getReportLine(MissingEditorialAttributeCheck.java:76)
>	at org.reactome.release.qa.check.MissingEditorialAttributeCheck.executeQACheck(MissingEditorialAttributeCheck.java:48)
>	at org.reactome.release.qa.Main.main(Main.java:140)

That's probably why I did it the other way. I think getAttributeValue is returning a `Boolean` in this case, not a String.

src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
SolomonShorser-OICR: I am not a fan of this change. It is an uncommon use of parentheses that I find makes the code weirder looking and harder to read.

src/main/java/org/reactome/release/qa/check/MissingEditorialAttributeCheck.java
jweiser: Ah okay - I thought that a String was returned.  In that case, the way you have it makes sense.

